Issue #7, opened 2019-10-11
Add configurable retries to all requests


marcphilipp:
close and release operations in Nexus can take some time. It is required to wait a little bit until the command is effectively performed (i.e. is not in the transitioning state) after it was accepted.

Issue #6, opened 2019-10-11
Add task for releasing a staging repository


marcphilipp:

ihostage:
Close/Release repository tasks must have ordering settings.
For details Codearte/gradle-nexus-staging-plugin#157

szpak:
@ihostage It's already [implemented[(https://github.com/gradle-nexus/publish-plugin/blob/master/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt#L148-L156) as part of [this PR](https://github.com/gradle-nexus/publish-plugin/pull/9).

I will update the tasks based on my recent development effort with releasing.

ihostage:
Thank a lot, @szpak 👍

szpak:
Implemented in #18.

```
close${repositoryName()}StagingRepository
release${repositoryName()}StagingRepository
```

Issue #4, opened 2019-10-11
Add task for closing a staging repository


marcphilipp:
- Allow close task to be run in a separate Gradle invocation. That requires that initialize task to write the necessary information (staging id?) to a file so the close task can read it from there
- Add `mustRunAfter` dependencies to all publish tasks (across projects) to support parallel multi-project builds

vlsi:
Note: "staging repository id" should be visible to the end-user anyway.
There should be a way to use the ide for other tasks.

Use case: generate draft mail on "Nexus repository preview"

szpak:
Yes, we plan to expose it as an extension property as [currently in GNSP.](https://github.com/Codearte/gradle-nexus-staging-plugin/blob/df4ee7187839b2ef978486950e75c0832948b0f0/src/main/groovy/io/codearte/gradle/nexus/NexusStagingExtension.groovy#L27)

vlsi:
Can it be a task property rather than an extension property?

szpak:
As I will be set by one task (`initialize*`) and consumed by others (`close*` and `release*`) for me it's better to keep it in a "global" space - the extension. However, having good argument against that could be a good input to rethink the idea. 
For example, a potential issue I see is the fact that the plugin will support multiple `initialize*` tasks for different Nexus servers which would complicate the extension. However, maybe we will have to have there some mapping server -> server properties anyone. Would it be enough for you? Is there any other reason that you would prefer to keep it in the task (and which task - init/close/release/drop/all of them?) instead?

vlsi:
> Is there any other reason that you would prefer to keep it in the task

The value is not usable before the task is executed, so it looks pretty natural to add `dependsOn(initialize)` + treat the id as the task output.

szpak:
> The value is not usable before the task is executed
Not always. You may want to close/release a staging repository created with some older Gralde execution (on demand or dealing with some errors).
It would be especially usable with `drop*` task (once/if implemented).

vlsi:
Do you mean the property would be "read only"?
That is user would never need to use `.set(..)`

If that is the case, it might be ok. However, having a single property would indeed impact the ability to have multiple repositories.

Then, Gradle "forbids" cross-project access. In other words, the order of project evaluation is not really consistent, so it might be the case that subproject is evaluated before the root project, thus the extension might be not there (or be empty).

AFAIK "it is a code smell every time cross-project APIs are invoked".
[...]

PR #9, opened 2019-10-18
[WIP] [#4][#6] Close and release tasks

szpak:
The initial implementation of close and release tasks. It's effectively broken with the real Nexus instance (due to missing retrier to handle "in transition" periods - #7), but I wanted to cover it in another PR as I had some issue with Kotlin and I'm not sure about some Gradle related bindings here. Therefore, I would like to get faster feedback for the following code.

In the next PR I plan to work on:
 - retrier to cover "inTransition == true" case
 - E2E tests
 - duplication reduction in tests
 - removing GNSP reference


src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -16,67 +16,23 @@
...
-        username.set(repository.username)
-        password.set(repository.password)
-        packageGroup.set(extension.packageGroup)
-        stagingProfileId.set(repository.stagingProfileId)
-        repositoryName.set(repository.name)
-        clientTimeout.set(extension.clientTimeout)
-        connectTimeout.set(extension.connectTimeout)
-        this.onlyIf { extension.useStaging.getOrElse(false) }
-    }
+    private val stagingRepositoryId: Property<String> = repository.stagingRepositoryId
```
szpak: I'm not sure about the best way to bind stagingRepositoryId in `InitializeNexusStagingRepository` and `Repository.stagingRepositoryId`, to be able to set value in task to set it in Extension/Repository. I initially just set `stagingRepositoryId: Property<String> = repository.stagingRepositoryId`, but I don't like sharing a reference, so maybe you have any better idea.

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
```diff
@@ -0,0 +1,109 @@
...
+    internal fun `close task should run after init`() {
+        initSingleProjectWithDefaultConfiguration()
+        assertGivenTaskMustRunAfterAnother("closeSonatypeStagingRepository", "initializeSonatypeStagingRepository")
+    }
+
+    @Test
+    @Disabled("Broken - must run only after publishMavenJavaPublicationToSonatypeRepository not publishToSonatypeRepository")
+    internal fun `close task should run after related publish`() {
+        initSingleProjectWithDefaultConfiguration()
+        assertGivenTaskMustRunAfterAnother("closeSonatypeStagingRepository", "publishToSonatypeRepository")
```
szpak: publishToNexusTask (and close/release tasks) depends/mustRunAfter publishMavenJavaPublicationToSonatypeRepository. It would be probably better to use publishToSonatypeRepository to also cover PDF documentation update, CDC stubs, etc. @marcphilipp How would be tweak it to cover that?

szpak:
Hmm, is it possible to switch to "[draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/)"?

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
vlsi: ` = objects.property().convention(repository.stagingRepositoryId)` ?

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -122,4 +69,8 @@ constructor(objects: ObjectFactory, extension: NexusPublishExtension, repository
         logger.info("Updating URL of publishing repository '{}' to '{}'", repository.name, url)
         repository.setUrl(url.toString())
     }
+
+    private fun keppStagingRepositoyIdInExtension(stagingRepositoryIdAsString: String) {
```
vlsi: kepp?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepository.kt
```diff
@@ -38,4 +38,10 @@ open class NexusRepository @Inject constructor(val name: String, project: Projec
     }
 
     val stagingProfileId: Property<String> = project.objects.property()
+
+    val stagingRepositoryId: Property<String> = project.objects.property()
+
+    fun capitalizedName(): String {
```
vlsi: should this be `internal`? (== avoid adding that to the public API)

src/main/kotlin/io/github/gradlenexus/publishplugin/ReleaseNexusStagingRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+@Suppress("UnstableApiUsage")
+open class ReleaseNexusStagingRepository @Inject
+constructor(objects: ObjectFactory, extension: NexusPublishExtension, repository: NexusRepository) :
+        BaseOperationOnNexusStagingRepository(objects, extension, repository) {
+
+    @get:Input
+    val stagingRepositoryId: Property<String> = objects.property()
+
+    init {
+        stagingRepositoryId.set(repository.stagingRepositoryId)
```
vlsi: Should this be `val stagingRepositoryId: Property<String> = objects.property().convention(repository.stagingRepositoryId)` ?

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
```diff
@@ -97,6 +97,28 @@ class NexusClient(private val baseUrl: URI, username: String?, password: String?
         }
     }
 
+    fun closeStagingRepository(stagingRepositoryId: String) {
+        try {
+            val response = api.closeStagingRepo(Dto(StagingRepositoryToTransit(listOf(stagingRepositoryId), "Closed by io.github.gradle-nexus.publish-plugin Gradle plugin"))).execute()
```
vlsi: I wish there was a more-or-less meaningful (configurable?) comment.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
```diff
@@ -97,6 +97,28 @@ class NexusClient(private val baseUrl: URI, username: String?, password: String?
...
+            if (!response.isSuccessful) {
+                throw failure("close staging repository", response)
```
vlsi: Should failure message include the repository id / group id?

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
```diff
@@ -97,6 +97,28 @@ class NexusClient(private val baseUrl: URI, username: String?, password: String?
...
+            }
+        } catch (e: IOException) {
+            throw UncheckedIOException(e)
```
vlsi: Is it required?

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
szpak: Please raise an issue for that. Could be implemented in the future - not an MVP.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
szpak: I plan to make error handling for general for all the calls.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
szpak: Hmm, it's most likely a leftover from the times the code was written in Java :).

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
szpak: Hmm, Unless I don't understand `convention()` properly, I will not be able to propagate value back to repository once it is set in the task. I need the other way round connection: task -> repository.

src/main/kotlin/io/github/gradlenexus/publishplugin/ReleaseNexusStagingRepository.kt
szpak: It could be. Assuming someone will want to set the value in a task directly (instead of playing with the `repository` structure).

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
vlsi: Oh, now I understand your intention.

Can you please clarify how do you mean the task should be used?
I think the user does not know `stagingRepositoryId` before the task is executed, so I don't see how the user would configure the property.

szpak: User could override the property in extension.repository to for example close or releate a requested staging repository with id known from the previous (failed) closing operation.

vlsi: What if I want to "discard old repository", "initialize new one", and "close it" in one go?

Use case: "produce next release candidate"
If old release candidate exists, then it might make sense to drop the previous candidate. Of course, the new candidate should be pushed to a different repository => need two different repository ids for the same nexus server within a same build.

Use case: "publish release and drop all release candidate repositories".
If release candidates are kept till the release is promoted (e.g. to enable comparison between RC artifacts), then "promote release" tasks needs to be able to drop multiple distinct repositories => again it is a case for "more than one staging id per nexus server within a single Gradle invocation".

szpak: How problematic would it be to do the clean up in the second call? E.g.
> gw release* -PyourID* && gw dropAllRepositories*
(assuming there will be that task one day)

While technically it is possible to call close/release/drop commands in Nexus with multiple repository IDs it's not a case with simple init/publish/close/release use case. On the other hand I can't find any sensible obstacles to use `ListProperty<String>` in the extension and have init to set just one element (there could be also some helper method to set just one).

vlsi: > How problematic would it be to do the clean up in the second call? 

Users are not always comfortable with storing passwords/keys, so "release / drop" might require the user to type the password (or provide it with environment/property).
So your example with `&&` would be much more complicated, as there are multiple credentials involved. In Apache case the credentials are for `dist.apache.org` (SVN), `repository.apache.org` (Nexus), `gitbox.apache.org` (Git) all of which participate in the release candidate/release publish.

Having a single command helps there.

Then there's "error recovery". If the actions are split into multiple commands, it makes it more complicated to "resume after failure". It looks like the user would have to analyze which command failed in order to retry or rollback.

build.gradle.kts
```diff
@@ -98,9 +98,13 @@ dependencies {
 }
 
 stutter {
+    isSparse = (findProperty("stutter.sparce")?.toString()?.toBoolean()) ?: true
     java(8) {
         compatibleRange("4.10")
     }
+    java(11) {
+        compatibleRange("4.10")
+    }
```
marcphilipp: ❓ Why do we need this?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -55,7 +58,7 @@ class NexusPublishPluginTests {
 
     private val gson = Gson()
 
-    private val gradleVersion = System.getProperty("compat.gradle.version")
+    private val gradleVersion = System.getProperty("compat.gradle.version") ?: GradleVersion.current().version
```
marcphilipp: ❓ This should always be set since we only run these via a `Test` task configured by Stutter or am I missing sth.?

src/main/kotlin/io/github/gradlenexus/publishplugin/BaseOperationOnNexusStagingRepository.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.Optional
+import org.gradle.kotlin.dsl.property
+import java.net.URI
+import java.time.Duration
+import javax.inject.Inject
+
+@Suppress("UnstableApiUsage")
+abstract class BaseOperationOnNexusStagingRepository @Inject
```
marcphilipp: ❓ How about `AbstractNexusStagingRepositoryTask`?

src/main/kotlin/io/github/gradlenexus/publishplugin/BaseOperationOnNexusStagingRepository.kt
```diff
@@ -0,0 +1,86 @@
...
+    protected val stagingProfileId: Property<String> = objects.property()
+
+    @get:Input
+    protected val repositoryName: Property<String> = objects.property()
+
+    @get:Internal
+    protected val clientTimeout: Property<Duration> = objects.property()
+
+    @get:Internal
+    protected val connectTimeout: Property<Duration> = objects.property()
```
marcphilipp: ❓ Maybe we should extract these properties into a `@Nested` `NexusConfiguration` object instead of using a base class?

build.gradle.kts
szpak: I was surprised at first, but to have the configuration in stutter applied it is required to call `gw stutterWriteLocks` instead of just making changes in the configuration (I spent some time trying to explain why sparse mode doesn't work, even it was explicitly enabled in the project configuration...
Nevertheless, it is in fact no longer needed as fallback java8.lock will be taken instead. I will revert it.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
szpak: I'm running those tests from Idea itself (not Gradle from Idea) and then it doesn't seem to be set.

marcphilipp: Understood, would delegating test execution in IDEA work for you or is there a reason not to use that?

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.gradle.kotlin.dsl.property
+import javax.inject.Inject
+
+@Suppress("UnstableApiUsage")
+open class CloseNexusStagingRepository @Inject
+constructor(objects: ObjectFactory, extension: NexusPublishExtension, repository: NexusRepository) :
+        BaseOperationOnNexusStagingRepository(objects, extension, repository) {
+
+    @get:Input
+    val stagingRepositoryId: Property<String> = objects.property()
```
marcphilipp: Would it make sense to allow passing this as an `@Option`, too?

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
marcphilipp: I'm afraid I don't understand your question. Could you please elaborate a bit more?

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
marcphilipp: Why do we need this property in this class at all? Wouldn't it suffice to set it in `NexusRepository`? In addition, I think we should write it to a file and read it from there in `NexusRepository` by default. That way, one could initialize and publish in the first Gradle invocation, do some manual testing, and close and release without passing the `stagingRepositoryId` manually. Thought?

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
szpak: Sure. How the easiest would be to change the plugin implementation to make `closeSonatypeStagingRepository` depends on `publishToSonatypeRepository` instead of all the related publications separately (e.g. `publishMavenJavaPublicationToSonatypeRepository`, `publishPdfDocumentationPublicationToSonatypeRepository`)?

It's not a big deal as a publication for PDF documents should be also covered by:
```
        val publishTasks = project.tasks
                .withType<PublishToMavenRepository>()
                .matching { it.repository == nexusRepository }
```

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
szpak: Hmm, what would be the behavior in that case? Close all repositories?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
szpak: Running test from Idea itself it faster and I try to use it by default (although Idea guys made it a little bit [harder](https://youtrack.jetbrains.com/issue/IDEA-223392)). While for the functional tests with TestKit it's negligible, I like to still have an option to do that (especially if just some minor changes are required for that).

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
szpak: Currently `NexusRepository` is only passed to the constructor to collect some required values. The reference is not kept. However, if you are fine with that I will refactor the code.

Regarding keeping it in the file I propose to postpone it to Milestone 2 - it's not MVP.

src/main/kotlin/io/github/gradlenexus/publishplugin/BaseOperationOnNexusStagingRepository.kt
szpak: One could want to have a separated timeout per repository. However, there still should be an easy way to set it globally (as in most of the cases it is enough - in fact most of the users (except @vlsi ;) ) will only use one repository per build).

src/main/kotlin/io/github/gradlenexus/publishplugin/ReleaseNexusStagingRepository.kt
szpak: Btw, `Property.convention()` was added in 5.1. Should we care about 4.10 users? Or having 6.0 on the doorstep it's not a problem?

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
szpak: Changed.

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
marcphilipp: No, the option would allow passing the stagingRepositoryId as a task option on the command line, e.g. `closeSonatypeStagingRepository --staging-repo-id=comacme-1234`

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
marcphilipp: Ok, fair enough.

src/main/kotlin/io/github/gradlenexus/publishplugin/ReleaseNexusStagingRepository.kt
marcphilipp: Let's stay compatible with 4.10 for now. That's the reason I didn't use `convention()` in the other places as well.

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
marcphilipp: Can we just add `closeTask.mustRunAfter(publishToNexusTask)` and the same for `releaseTask`?

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
szpak: Sure, I confused `Option` with `Optional` :man_facepalming: 

`Option` is definitely a way to go :).

szpak: Done.

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
szpak: How would be best to get a reference/provider to the umbrella publish task in the production plugin code?

src/main/kotlin/io/github/gradlenexus/publishplugin/AbstractNexusStagingRepositoryTask.kt
```diff
@@ -0,0 +1,93 @@
...
+    @get:Input
+    protected val repositoryName: Property<String> = objects.property()
+
+    @get:Internal
+    protected val clientTimeout: Property<Duration> = objects.property()
+
+    @get:Internal
+    protected val connectTimeout: Property<Duration> = objects.property()
+
+    private val repository: Property<NexusRepository> = objects.property()
```
szpak: @marcphilipp Have you thought about something like to that to keep `NexusRepository` in a task? Our previous conversation disappeared :-/

marcphilipp: Well, we could use `@Nested` and remove all the fields that just copy its properties. Doing both looks redundant to me.

szpak:
Applied @marcphilipp changes from #11.

szpak:
Merging conflicts due to the recent changes (spotless upgrade and import reordering) in master :-/. Will fix.

marcphilipp:
Sorry! 😬

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -40,7 +39,6 @@ constructor(
     private val stagingRepositoryId: (String) -> Unit
 ) : AbstractNexusStagingRepositoryTask(objects, extension, repository) {
 
-    @Optional
```
marcphilipp: It's only required if `stagingProfileId` is not set. Thus, I think `@Optional` was correct, wasn't it?

szpak: Well, they both can be not set and then there will be `val packageGroup = packageGroup.get()` called. However, it would be problematic to add two-field present validation, so probably `packageGroup.get()` is good enough, especially as it reports the name of the not provided property. I will restore `@Optional`.

PR #11, opened 2020-01-11
Don't share state via NexusRepository

marcphilipp:
- Remove stagingProfileId from close/release tasks as it is not needed
- Wire stagingRepositoryId for close/release task in plugin
- Remove mutable state from NexusRepository
- Rename CLI option to --staging-repository-id to make it more idiomatic


src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -18,18 +18,33 @@ package io.github.gradlenexus.publishplugin
...
+    objects: ObjectFactory,
+    extension: NexusPublishExtension,
+    repository: NexusRepository,
+    private val serverUrlToStagingRepoUrl: MutableMap<URI, URI>,
+    private val stagingRepositoryId: (String) -> Unit
+) : AbstractNexusStagingRepositoryTask(objects, extension, repository) {
+
+    @Optional
+    @Input
+    val packageGroup = objects.property<String>().apply {
```
szpak: +1, `packageGroup` is not important for close and release.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -55,16 +56,23 @@ class NexusPublishPlugin : Plugin<Project> {
...
 
         extension.repositories.all {
+            val stagingRepositoryId = project.objects.property<String>()
             project.tasks.register("publishTo${capitalizedName()}") {
                 description = "Publishes all Maven publications produced by this project to the '${this@all.name}' Nexus repository."
                 group = PublishingPlugin.PUBLISH_TASK_GROUP
             }
             project.tasks
-                    .register<InitializeNexusStagingRepository>("initialize${capitalizedName()}StagingRepository", project.objects, extension, this, serverUrlToStagingRepoUrl)
+                    .register<InitializeNexusStagingRepository>("initialize${capitalizedName()}StagingRepository", project.objects, extension, this, serverUrlToStagingRepoUrl, { id: String -> stagingRepositoryId.set(id) })
```
szpak: I like the trick to use a function to set (and later use) dangling property in not directly bound contexts. It greatly reduced complexity of the workaround on finalized properties in Gradle.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepository.kt
```diff
@@ -17,46 +17,36 @@
...
+open class NexusRepository @Inject constructor(@Input val name: String, project: Project) {
 
-    @get:Input
-    val nexusUrl: Property<URI> = project.objects.property()
+    @Input
+    val nexusUrl = project.objects.property<URI>()
 
-    @get:Input
-    val snapshotRepositoryUrl: Property<URI> = project.objects.property()
+    @Input
```
szpak: Tests confirms it works, but the related [documentation](https://docs.gradle.org/6.0.1/javadoc/org/gradle/api/tasks/Input.html) doesn't cover that topic:
> This annotation should be attached to the getter method in Java or the property in Groovy. Annotations on setters or just the field in Java are ignored.

Is it the recommenced approach for Kotlin?  Then documentation could be improved to: `or the property in Groovy/Kotlin`.

marcphilipp: I'll bring this up with the Gradle team.

marcphilipp: It's now an issue: https://github.com/gradle/gradle/issues/11927

PR #12, opened 2020-01-12
Use Autostyle for formatting

vlsi:
See https://github.com/autostyle/autostyle/blob/master/CHANGES.md#version-30

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,10 +1,26 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     `kotlin-dsl`
     `maven-publish`
     id("com.gradle.plugin-publish") version "0.10.1"
-    id("com.diffplug.gradle.spotless") version "3.27.0"
+    id("com.github.autostyle") version "3.0"
     id("com.github.johnrengelman.shadow") version "5.2.0"
     id("org.jetbrains.gradle.plugin.idea-ext")
     id("com.github.ben-manes.versions") version "0.27.0"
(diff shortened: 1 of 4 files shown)


marcphilipp:
@vlsi What was the reason for forking Spotless?

vlsi:
Contributing to Spotless is hard :(
It seems to take more time than maintaining a fork.

I did try to contribute a feature: https://github.com/diffplug/spotless/pull/485
Unfortunately, the PR feedback looked like a bikeshedding to me :(

Spotless is written in Java, and their choice for multi-line strings in tests is

```java
		assertTaskFailure(task,
				"    testFile",
				"        @@ -2,9 +2,9 @@",
				"         u 1",
				"         u 2",
				"         u 3",
```

which is very unpleasant to edit and update :(

The PR was merged, however, [then I found a bug in my implementation](https://github.com/diffplug/spotless/pull/485#issuecomment-552925932):

> Vladimir: current.getEndA() - next.getBeginA() should flipped like next.getBeginA() - current.getEndA()

Ned has chosen to revert the PR altogether.

Just in case: the bug was not detected by Spotless test suite since none of the tests included a case when a file had multiple violations that were separated with more than 3 lines.

---

Ned is nice, however, their priorities seem to be tangential to my needs.

For instance, Spotless does not support license headers for `kts`, `xml`, `sh` files, it does not support `package-info.java`, it often produces "stacktrace-like" messages (e.g. unable to parse ...) where a single pointer to the line and position were enough.

For instance, there's an open issue https://github.com/diffplug/spotless/issues/496.
As you can see the proposed solution there is to add `$FIRST_LINE` to the license header.

On the other hand, I reworked that in Autostyle, so the license text is declared just once, and Autostyle formats it using the appropriate comment styles (e.g. it knows Java-like, XML, Shell, Windows CMD files).

---

Spotless uses obscure Gradle techniques, so you can't re-execute the task that failed (you can't really copy-paste the task name from the failure log).
It uses obscure techniques to configure the fileset, and it does not really support multi-project layouts (see https://github.com/diffplug/spotless/issues/468)
It used to produce Gradle compatibility warnings, and it seems still rely on `afterEvaluate` which is sad for multi-module projects.

It does not support Gradle Build Cache.

---

Have you seen "paddedcell warning" in Spotless?
It happens in case there's a runaway rule (e.g. a rule that adds a space to the end of the line on each execution).

When that happens Spotless provides no output on the actual contents, so you can't really understand what is misbehaving.

---

For instance, I've migrated Apache Calcite from Spotless to Autostyle: https://github.com/apache/calcite/pull/1682, Apache Calcite Avatica: https://github.com/apache/calcite-avatica/pull/118, Apache JMeter: https://github.com/apache/jmeter/pull/549.

It increased the consistency of the source code, and it improved the error messages.

---

All of the above (+ migration to Kotlin + migration to JUnit 5) seemed to be easier to implement in Autostyle than discussing in Spotless issues :(

marcphilipp:
@vlsi Thanks for the detailed explanation! I'll take a closer look.

marcphilipp:
@vlsi No offence, I think we'll stay with Spotless for now which works well enough for what we need for this small project here.

PR #13, opened 2020-01-12
Fix .gitignore: build -> /build

vlsi:
'build' pattern in .gitignore would exclude build folders in all the subfolders.
It might cause unexpected exclusion of files under packages
like io.github.gradlenexus.build...


szpak:
I had the same change in my local build with the intention to commit it later on.

PR #18, opened 2020-02-05
[#7] Configurable retries to close/release requests

### LINKED ISSUE #7: Add configurable retries to all requests

szpak:
his PR adds retrying on the close and release requests. It is required as Nexus only acknowledge that a given repository has been accepted for transition. With Sonatype Nexus is can take minutes to finish the operation.

I started with a basic external [retrier](https://github.com/alexo/retrier), but separated it to be able to switch to some other (external or in-house implementation if needed or desired).

I also considered this [retrier](https://github.com/ravichaturvedi/retrier) (it seems to only support throwing exceptions to indicate a not successful attempt) and that one for [kotlin](https://github.com/michaelbull/kotlin-retry) (it uses suspended function and I'm not sure how Gradle deals with that in the plugins and in addition it is not in Maven Central nor jCenter). In GNSP, I have a custom implementation (which among others nicely displays subsequent attempts), but I don't know if we want to reinvent a wheel here and maybe some external retrier will be fine.

I was able to successfully release locally a small sample project with the plugin as e2e tests. However, it requires some more work to be able to use it as automatic e2e tests from the CI server (#5). Then it would be good to split also functional tests into smaller pieces.

Btw, I moved MockServer to a field to simplify private method calls - it is used by (almost) everything in that class. It makes a diff less readable. I may revert it and propose as a separate PR, if you prefer.

Closes #7.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -92,6 +92,8 @@ dependencies {
     testImplementation("com.github.tomakehurst:wiremock:2.25.1")
     testImplementation("ru.lanwen.wiremock:wiremock-junit5:1.3.1")
     testImplementation("org.assertj:assertj-core:3.14.0")
+    testImplementation("org.mockito:mockito-junit-jupiter:3.2.4")
+    testImplementation("com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0")
 }
 
 stutter {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,9 +80,7 @@ configurations {
 dependencies {
     shadowed("com.squareup.retrofit2:retrofit:2.7.1")
     shadowed("com.squareup.retrofit2:converter-gson:2.7.1")
-    shadowed("io.github.alexo:retrier:0.1") {
-        exclude("org.mockito")  // https://github.com/alexo/retrier/pull/1
-    }
+    shadowed("net.jodah:failsafe:2.3.3")
 
     val nexusStagingPlugin = create("io.codearte.gradle.nexus:gradle-nexus-staging-plugin:0.21.2")
     compileOnly(nexusStagingPlugin)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,6 +80,9 @@ configurations {
 dependencies {
     shadowed("com.squareup.retrofit2:retrofit:2.7.1")
     shadowed("com.squareup.retrofit2:converter-gson:2.7.1")
+    shadowed("io.github.alexo:retrier:0.1") {
+        exclude("org.mockito")  // https://github.com/alexo/retrier/pull/1
+    }
 
     val nexusStagingPlugin = create("io.codearte.gradle.nexus:gradle-nexus-staging-plugin:0.21.2")
     compileOnly(nexusStagingPlugin)
@@ -110,7 +113,9 @@ configurations {
 sourceSets {
     compatTest {
         compileClasspath += sourceSets["test"].output
+        compileClasspath += sourceSets["main"].output
         runtimeClasspath += sourceSets["test"].output
+        runtimeClasspath += sourceSets["main"].output
     }
 }
 
--- a/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
+++ b/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
@@ -678,8 +678,7 @@ class NexusPublishPluginTests {
                 .withRequestBody(matchingJsonPath("\$.data[?(@.stagedRepositoryIds[0] == '$stagingRepositoryId')]"))
                 .withRequestBody(matchingJsonPath("\$.data[?(@.autoDropAfterRelease == true)]"))
                 .willReturn(aResponse().withHeader("Content-Type", "application/json").withBody("{}")))
-        server.stubFor(get(urlEqualTo("/staging/repository/$stagingRepositoryId"))
-                .willReturn(aResponse().withHeader("Content-Type", "application/json").withBody("{\"transitioning\":false,\"type\":\"${operation.type}\"}")))
+        stubGetStagingRepoWithIdAndStateRequest(server, stagingRepositoryId, operation.type)
     }
 
     @Test
@@ -837,6 +836,15 @@ class NexusPublishPluginTests {
                             .withBody(getOneStagingProfileWithGivenIdShrunkJsonResponseAsString(stagingProfileId))))
     }
 
+    private fun stubGetStagingRepoWithIdAndStateRequest(server: WireMockServer, stagedRepositoryId: String, repoState: String) {
+        server.stubFor(get(urlEqualTo("/staging/repository/$stagedRepositoryId"))
+                        .withHeader("Accept", containing("application/json"))
+                        .willReturn(aResponse()
+                            .withStatus(200)
+                            .withHeader("Content-Type", "application/json")
+                            .withBody(getOneStagingRepoWithGivenIdJsonResponseAsString(stagedRepositoryId, repoState))))
+    }
+
     private fun expectArtifactUploads(server: WireMockServer, prefix: String) {
         server.stubFor(put(urlMatching("$prefix/.+"))
                 .willReturn(aResponse().withStatus(201)))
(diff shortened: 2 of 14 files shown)


build.gradle.kts
```diff
@@ -110,7 +115,9 @@ configurations {
 sourceSets {
     compatTest {
         compileClasspath += sourceSets["test"].output
+        compileClasspath += sourceSets["main"].output
```
vlsi: Do you mean `.output.classesDirs` here?

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/ActionRetrier.kt
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.gradlenexus.publishplugin.internal
+
+interface ActionRetrier<R> {
+    fun execute(operationToExecuteWithRetrying: () -> R): R
+}
```
vlsi: Kotlin does not support implementation of `Kotlin` interfaces via lambdas.

Do you think it would be better to use `typealias ActionRetrier`?

vlsi: Ah, `typealias` can't have generics :(

szpak: > Kotlin does not support implementation of `Kotlin` interfaces via lambdas.

I didn't know. I could change it to Java interface if that helped.

> Do you think it would be better to use `typealias ActionRetrier`?

`typealias` seems to be more an a type alias (like in Groovy or Scala) and I would like to have an interface with (potentially) multiple implementations.

build.gradle.kts
szpak: The above works. Probably there is an overloaded method to support it.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/ActionRetrier.kt
vlsi: >I could change it to Java interface if that helped.

In fact, it would help. In other words, Kotlin can implement Java interfaces via lambdas.
However, it is a bit weird for me :(

build.gradle.kts
vlsi: `.output` is more than `classes`. it includes resources as well.

Do you really need to include `resources` on the **compile**  classpath?

szpak: To compile not, but in runtime it might be necessary in some cases. E.g. with Nebula Test it would be required to provide a properties file with the main plugin class provided (with a plugin for plugins and TestKit is not necessary, however).

build.gradle.kts
```diff
@@ -80,6 +80,9 @@ configurations {
 dependencies {
     shadowed("com.squareup.retrofit2:retrofit:2.7.1")
     shadowed("com.squareup.retrofit2:converter-gson:2.7.1")
+    shadowed("io.github.alexo:retrier:0.1") {
+        exclude("org.mockito")  // https://github.com/alexo/retrier/pull/1
+    }
```
marcphilipp: I've had a good experience with https://jodah.net/failsafe/ for this in the past. It seems more mature, too. Should we switch?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepository.kt
```diff
@@ -48,6 +52,18 @@ open class NexusRepository @Inject constructor(@Input val name: String, project:
...
 
+    @Nested
+    val retrying = project.objects.property<RetryingConfig>().apply {
+        set(RetryingConfig(project))
+    }
+
+    fun retrying(action: Action<in RetryingConfig>) = action.execute(retrying.get())
+
+    //TODO: Fails anyway with Groovy for "maxNumber = 10" with "Cannot cast object '10' with class 'java.lang.Integer' to class 'org.gradle.api.provider.Property'"
+    fun retrying(closure: Closure<*>) {
```
marcphilipp: :x: We don't need this method because you can write an `Action` as a closure in Groovy.

szpak: I believe that there was something mentioned in the Gradle documentation that Gradle doesn't support the second level of nesting domain objects and it is needed. Unfortunately, I cannot find it right now. Nevertheless, it doesn't work as mentioned in the comment for Groovy anyway, so I maybe remove it anyway.

build.gradle.kts
szpak: I didn't know it. I will take a look.

szpak: It's very nice. That's pity that is was in my search results at the end of the second page :-/. Replaced.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/BasicActionRetrier.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.github.gradlenexus.publishplugin.StagingRepository
+import net.jodah.failsafe.Failsafe
+import net.jodah.failsafe.RetryPolicy
+
+open class BasicActionRetrier<R>(retryingConfig: RetryingConfig, stopFunction: (R) -> Boolean) : ActionRetrier<R> {
+
+    private val retrier: RetryPolicy<R> = RetryPolicy<R>()
+            .handleResultIf(stopFunction)
+            .withMaxRetries(retryingConfig.maxRetries.get())
+            .withDelay(retryingConfig.delayBetween.get())
```
szpak: Failsafe is nice and probably well tested, but I would like to test that I pass the parameters correctly. Functional tests check retrying in general (if it occurs), but I would like to test some corner cases (such as a reaction on particular exceptions) at the unit (testing) layer.
Unfortunately, `RetryingConfig` relies on the Gradle API (for `Property`). I initially modeled it as a implementation of a generic interface which just provides values, but it complicated the code. Do you have any better/simpler idea how would we make `BasicActionRetries` less Gradle dependent?

marcphilipp: I think I would pass in `retryingConfig.maxRetries.get()` and `retryingConfig.delayBetween.get()` to the constructor instead of `RetryingConfig`.

szpak: That's definitely much simpler solution :-). Probably we will not have (much) more parameters passes here in the future.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/ActionRetrier.kt
szpak: I gave it a try, but it complicates the current implementation a little bit.

```
interface ActionRetrier<R> {
    R execute(Supplier<R> operationToExecuteWithRetrying);
}
```
cannot be used as FailSafe which we use expects its own:
```
public interface CheckedSupplier<T> {
  T get() throws Exception;
}
```
Callable would be ok, but at the type level they are not compatible. As a result:
```
    override fun execute(operationToExecuteWithRetrying: Callable<R>): R {
        return Failsafe.with(retrier).get(
            { operationToExecuteWithRetrying.call() } as CheckedSupplier<R>) }
        )
```
is required which doesn't look good. 

I will merge it as is (with the Kotlin interface). Feel free to create a separate PR to improve that.

src/main/kotlin/io/github/gradlenexus/publishplugin/RetryingConfig.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    @Internal
+    val maxRetries = project.objects.property<Int>().apply {
+        set(DEFAULT_MAXIMUM_NUMBER_OF_RETRIES)
+    }
+
+    @Internal
+    val delayBetween = project.objects.property<Duration>().apply {
+        set(DEFAULT_DELAY_BETWEEN_RETRIES)
+    }
```
vlsi: The naming is misleading.
`RetryingConfig` + `maxRetries` means the plugin would **retry** publication multiple times, however, in practice it means it would "just wait the specified number of times while the repository is in progress".

Do you think it would be better to use `timeout`-like configuration properties?

For instance `connectToNexusTimeout`, `initializeRepositoryTimeout`, `singleArtifactPublishTimeout`, `closeRepositoryTimeout`, etc, etc.

Note: the users do not care what the number of "status check" attempts is there. What they care is **ovreall timeout value**, and they might want to have something like `initial delay between status check attempts`.


WDYT?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -497,6 +501,7 @@ class NexusPublishPluginTests {
     }
 
     @Test
+    @Disabled("Fails on my Fedora...")
```
szpak: Ups, I have to do some conditional checking to exclude only my system as it's hard to keep it locally and do not commit accidentally...

PR #21, opened 2020-05-16
E2E tests with Sonatype Nexus

### LINKED ISSUE #5: Set up end-to-end tests with OSSRH Nexus

szpak:
I finally sat down and finished E2E testing with Sonatype Nexus (TestKit is lightly different in behaviors than nebula-test which I used before). There is are artifacts of the [minimal project published to Nexus](https://github.com/gradle-nexus-e2e/nexus-publish-e2e-minimal), the repository is closed and released. It can detect some regressions in the communications with Nexus (or occasionally fail on timeout ;-) ).

There is a [dedicated flow](https://github.com/gradle-nexus/publish-plugin/actions/runs/106832486) in GitHub Actions which executes "unit" and e2e tests (without the Gradle version compatibility tests). I needed to script it slightly, maybe it can be done easier in GH Actions.

The flow on the Sonatype side is constructed that way that on the staging repository close the artifacts are pushed into `/dev/null` instead of the real Maven Central. It should be close to production (however, as it turned out with [MVNCENTRAL-5276](https://issues.sonatype.org/browse/MVNCENTRAL-5276), that "fake" repo doesn't fails on superficial signature files).

A separate Nexus user it used which can't release the plugin itself (just that e2e project). Also the separate set of properties (ending with `E2E`) is used to do not accidentally touch production credentials while running locally. As TestKit hides `~/.gradle/gradle.properties` (nice!), I pass the variables to tests itself. The minimal project can be also executed directly from the console (with `*E2E` properties set locally). Probably it would be good to review the minimal project where most likely something could be simplified.

Fixes #5.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,31 @@
+name: E2E tests
+
+on:
+  push:
+    branches:
+      - master
+      - "szpak/*"
+      - "marc/*"
+
+jobs:
+  gradle:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout project with submodules
+      uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Setup JDK 8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    - name: Run sanity check
+      run: ./gradlew --scan --stacktrace test
+    - name: Run E2E tests
+      env:
+        ORG_GRADLE_PROJECT_sonatypeUsernameE2E: ${{ secrets.SONATYPE_USERNAME_E2E }}
+        ORG_GRADLE_PROJECT_sonatypePasswordE2E: ${{ secrets.SONATYPE_PASSWORD_E2E }}
+        ORG_GRADLE_PROJECT_signingKeyE2E: ${{ secrets.GPG_SIGNING_KEY_E2E }}
+        ORG_GRADLE_PROJECT_signingPasswordE2E: ${{ secrets.GPG_SIGNING_KEY_PASSPHRASE_E2E }}
+      run: |
+        ./gradlew --stacktrace --info e2eTest
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "src/e2eTest/resources/nexus-publish-e2e-minimal"]
 	path = src/e2eTest/resources/nexus-publish-e2e-minimal
-	url = git@github.com:gradle-nexus-e2e/nexus-publish-e2e-minimal.git
+	url = git@github.com:gradle-nexus/nexus-publish-e2e-minimal.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/e2eTest/resources/nexus-publish-e2e-minimal"]
+	path = src/e2eTest/resources/nexus-publish-e2e-minimal
+	url = git@github.com:gradle-nexus-e2e/nexus-publish-e2e-minimal.git
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -104,6 +104,13 @@ stutter {
     }
 }
 
+val e2eTest by sourceSets.creating {    //separate infrastructure as compatTest is called multiple times with different Java versions
+    compileClasspath += sourceSets["compatTest"].output
+    compileClasspath += sourceSets["main"].output
+    runtimeClasspath += sourceSets["compatTest"].output
+    runtimeClasspath += sourceSets["main"].output
+}
+
 configurations {
     compatTestCompileClasspath {
         extendsFrom(testCompileClasspath.get())
@@ -111,10 +118,10 @@ configurations {
     compatTestRuntimeClasspath {
         extendsFrom(testRuntimeClasspath.get())
     }
-    create("e2eTestCompileClasspath") {
+    configurations.named(e2eTest.implementationConfigurationName) {
         extendsFrom(compatTestCompileClasspath.get())
     }
-    create("e2eTestRuntimeClasspath") {
+    configurations.named(e2eTest.runtimeOnlyConfigurationName) {
         extendsFrom(compatTestRuntimeClasspath.get())
     }
 }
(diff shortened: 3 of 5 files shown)


build.gradle.kts
```diff
@@ -111,6 +111,12 @@ configurations {
     compatTestRuntimeClasspath {
         extendsFrom(testRuntimeClasspath.get())
     }
+    create("e2eTestCompileClasspath") {
```
szpak, line 114: Can I generate something in Kotlin to be able to refer to them by name?

build.gradle.kts
```diff
@@ -142,6 +154,16 @@ tasks {
         pluginClasspath.from.clear()
         pluginClasspath.from(shadowJar)
     }
+    register<Test>("e2eTest") {
+        testClassesDirs = sourceSets.get("e2eTest").output.classesDirs
+        classpath = sourceSets.get("e2eTest").runtimeClasspath
```
szpak, line 159: `sourceSets.e2eTest.runtimeClasspath` known from Groovy would be more readable here.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/BaseGradleTest.kt
```diff
@@ -0,0 +1,56 @@
...
+package io.github.gradlenexus.publishplugin
+
+import org.assertj.core.api.Assertions
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.jupiter.api.io.TempDir
+import java.nio.file.Path
+
+abstract class BaseGradleTest {
```
szpak, line 26: In the next iteration that class with be reused also in compatibility tests with Wiremock.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/e2eTest/resources/nexus-publish-e2e-minimal"]
+	path = src/e2eTest/resources/nexus-publish-e2e-minimal
+	url = git@github.com:gradle-nexus-e2e/nexus-publish-e2e-minimal.git
```
marcphilipp, line 3: What are the benefits of having this in a separate repo in a separate organization?

build.gradle.kts
```diff
@@ -142,6 +154,18 @@ tasks {
         pluginClasspath.from.clear()
         pluginClasspath.from(shadowJar)
     }
+    register<Test>("e2eTest") {
+        description = "Run E2E tests."
+        group = "Verification"
+        testClassesDirs = sourceSets.get("e2eTest").output.classesDirs
+        classpath = sourceSets.get("e2eTest").runtimeClasspath
```
marcphilipp, line 161:
```suggestion
        testClassesDirs = e2eTest.output.classesDirs
        classpath = e2eTest.runtimeClasspath
```

build.gradle.kts
```diff
@@ -120,6 +126,12 @@ sourceSets {
         runtimeClasspath += sourceSets["test"].output
         runtimeClasspath += sourceSets["main"].output
     }
+    create("e2eTest") {     //separate infrastructure as compatTest is called multiple times with different Java versions
+        compileClasspath += sourceSets["compatTest"].output
+        compileClasspath += sourceSets["main"].output
+        runtimeClasspath += sourceSets["compatTest"].output
+        runtimeClasspath += sourceSets["main"].output
+    }
```
marcphilipp, line 134: If you create the source set outside of the `sourceSet` "block", you can reference it below.


```kotlin
val e2eTest by sourceSets.creating {
    compileClasspath += sourceSets["compatTest"].output
    compileClasspath += sourceSets["main"].output
    runtimeClasspath += sourceSets["compatTest"].output
    runtimeClasspath += sourceSets["main"].output
}
```

build.gradle.kts
marcphilipp: You shouldn't have to create these configurations since the source set will already create them for you. After creating the source set, you can configure them like this:

```kotlin
configurations.named(e2eTest.implementationConfigurationName) {
    extendsFrom(configurations.testImplementation.get())
}

configurations.named(e2eTest.runtimeOnlyConfigurationName) {
    extendsFrom(configurations.testRuntimeOnly.get())
}
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,40 @@
...
+      env:
+        GPG_SIGNING_KEY_ID: 6528359BBE836F20
+        GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
+        GPG_SIGNING_KEY_PASSPHRASE: ${{ secrets.GPG_SIGNING_KEY_PASSPHRASE }}
+      run: |
+        echo "$GPG_SIGNING_KEY" | gpg --fast-import --no-tty --batch --yes --passphrase $GPG_SIGNING_KEY_PASSPHRASE
+        # Environment variables cannot have "." and it's problematic to pass them with "ORG_GRADLE_PROJECT_"
+        mkdir -p $HOME/.gradle
+        echo "signing.gnupg.keyNameE2E=$GPG_SIGNING_KEY_ID" >> ~/.gradle/gradle.properties
+        echo "signing.gnupg.passphraseE2E=$GPG_SIGNING_KEY_PASSPHRASE" >> ~/.gradle/gradle.properties
```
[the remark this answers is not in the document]
szpak: Nice! I didn't know about that feature, it simplifies that kind of cases greatly.

.gitmodules
szpak: A valid question. I initially created it just in the subpackage of `io.github.gradle-nexus` - `io.github.gradle-nexus.e2e`. However, doing the Nexus stuff I realized that it causes the staging profiles overlapping. As a result, using the production staging profile for `io.github.gradle-nexus` would allow releasing also e2e project (it applies also to all subpackages). However, in Nexus the release repository is bound to the staging profile (the second column from right).

image

As a result, it would be possible to accidentally release the e2e project to Maven Central, using my or your production credentials. Having a separate group - `io.github.gradle-nexus-e2e` (vs `io.github.gradle-nexus`) prevents that (the "No Sync Releases repo is always used).

I needed a separate organization to cover a domain requirement in Sonatype rules. However, after creating that PR, I've got a flash of inspiration that it is not a problem to move that e2e project back to the main organization (keeping just the empty organization in place for the domain requirement). I will update that, keeping the old `groupId` for the aforementioned reasons.

.github/workflows/e2e.yml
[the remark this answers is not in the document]
marcphilipp: Well, I implemented it so I knew about it. 😉

PR #22, opened 2020-05-17
Use BuildService on Gradle 6.1+

marcphilipp:
In order to support Gradle's upcoming configuration cache feature
(formerly known as "instant execution"), we no longer use static state
and a build listener to clear it, but rely on a BuildService instead.


marcphilipp:
Superseded by #28

Issue #25, opened 2020-05-18
Sanity tests with other supported Java versions


szpak:
Using the CI server. Probably 8, 11 and 14 at the moment.

JUnit 5 is tested with different Java versions (including EA). Maybe it would be possible/feasible to leverage their infrastructure.

PR #26, opened 2020-05-18
[#25] CI sanity testing with different Java versions

szpak:
8, 11, 14.

Fixes: #25.

### DIFF
--- a/.github/workflows/java-versions.yml
+++ b/.github/workflows/java-versions.yml
@@ -1,0 +1,36 @@
+name: Java cross-version tests
+
+on:
+  push:
+    branches:
+      - master
+      - "szpak/*"
+      - "marc/*"
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  openjdk14:
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/java-versions.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          java-version: ${{ matrix.jdk }}
+      # Workaround https://github.com/ajoberstar/gradle-stutter/issues/22
+      - name: Reduce number of Gradle regressions builds
+        run: |
+          for l in .stutter/*.lock; do
+            echo $l
+            echo "$(cat $l | grep -v "^#" | tail -1)" > $l
+            cat $l
+          done
```
marcphilipp, line 32: How about just calling one of the compat test tasks or settiong a project property that disables all but one?

szpak: That would be best and I asked about that once I've got into the project in October:
https://github.com/ajoberstar/gradle-stutter/issues/22

If you know an easy way to tweak it at the Gradle level, we definitely can switch to it.

PR #27, opened 2020-05-20
[#5] Prefer envVars when passing properties to E2E test

szpak:
To do not display their values with --info (as parameters passed to
Gradle executor. Properties with "." cannot be passed that way.
They should be avoided on a CI server.


szpak:
I accidentally merged locally into master not rebased branch... Fixed by an extra merge.

PR #28, opened 2020-05-21
Require plugin to be applied on root project

marcphilipp:
The plugin now requires to be applied to the root project to ensure
that each Nexus repository is only declared once. It then adds a Maven
repository for publishing to the root and all subprojects that have the
maven-publish plugin applied.


src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -40,29 +34,60 @@ import org.gradle.kotlin.dsl.withType
...
+                                    .named("release${nexusRepo.capitalizedName()}StagingRepository")
+                            configureTaskDependencies(publishingProject, initializeTask, closeTask, releaseTask, mavenRepo)
+                        }
+                    }
+                }
+            }
+        }
+    }
 
+    private fun configureTasks(project: Project, extension: NexusPublishExtension) {
```
szpak, line 86: Just the nitpicking, but I would put that method before `configureTaskDependenciesForAllProjects` (it is called first in `apply()`).

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -40,29 +34,60 @@ import org.gradle.kotlin.dsl.withType
...
+                username.set(project.provider { nexusStagingExtension.username })
+                password.set(project.provider { nexusStagingExtension.password })
             }
-        })
+        }
+    }
 
-        val extension = project.extensions.create<NexusPublishExtension>(NexusPublishExtension.NAME, project)
+    private fun configureTaskDependenciesForAllProjects(rootProject: Project, extension: NexusPublishExtension) {
+        rootProject.afterEvaluate {
```
vlsi, line 65: I guess this `afterEvaluate` is not needed

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -40,29 +34,60 @@ import org.gradle.kotlin.dsl.withType
...
+            rootProject.allprojects {
+                val publishingProject = this
+                pluginManager.withPlugin("maven-publish") {
+                    publishingProject.afterEvaluate {
```
vlsi, line 69: `configureTaskDependencies` is using `afterEvaluate` internally.
I guess it would be better to refrain from `afterEvaluate { afterEvaluate {..}}` code patterns

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -40,29 +34,60 @@ import org.gradle.kotlin.dsl.withType
...
 class NexusPublishPlugin : Plugin<Project> {
 
-    companion object {
-        private val serverUrlToStagingRepoUrl = ConcurrentHashMap<URI, URI>()
-    }
-
     override fun apply(project: Project) {
-        project.pluginManager.apply(MavenPublishPlugin::class)
+        require(project == project.rootProject) {
+            "Plugin must be applied to the root project"
```
vlsi, line 39: It would probably make sense to add the actual project path to the error message.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -40,29 +34,60 @@ import org.gradle.kotlin.dsl.withType
...
+                pluginManager.withPlugin("maven-publish") {
+                    publishingProject.afterEvaluate {
+                        val nexusRepositories = addMavenRepositories(publishingProject, extension)
+                        nexusRepositories.forEach { (nexusRepo, mavenRepo) ->
+                            val initializeTask = rootProject.tasks.withType(InitializeNexusStagingRepository::class)
+                                    .named("initialize${nexusRepo.capitalizedName()}StagingRepository")
+                            val closeTask = rootProject.tasks.withType(CloseNexusStagingRepository::class)
+                                    .named("close${nexusRepo.capitalizedName()}StagingRepository")
+                            val releaseTask = rootProject.tasks.withType(ReleaseNexusStagingRepository::class)
+                                    .named("release${nexusRepo.capitalizedName()}StagingRepository")
```
vlsi, line 77: This looks like a loop invariant. Should it be moved out of the loop? (before `allprojects`)

It would probably make sense to use a single `named` call rather than `withType + named`:

```kotlin
rootProject.tasks.named<InitializeNexusStagingRepository>(
    "initialize${nexusRepo.capitalizedName()}StagingRepository"
)
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -40,29 +34,60 @@ import org.gradle.kotlin.dsl.withType
...
 
-        project.gradle.addBuildListener(object : BuildAdapter() {
-            override fun buildFinished(result: BuildResult) {
-                serverUrlToStagingRepoUrl.clear()
+        val extension = project.extensions.create<NexusPublishExtension>(NexusPublishExtension.NAME, project)
+        configureTasks(project, extension)
+        configureTaskDependenciesForAllProjects(project, extension)
+
+        project.rootProject.plugins.withId("io.codearte.nexus-staging") {
+            val nexusStagingExtension = project.rootProject.the<NexusStagingExtension>()
```
vlsi, line 47: should this be removed?

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -88,9 +87,13 @@ constructor(
...
-        val publishing = project.the<PublishingExtension>()
-        val repository = publishing.repositories.getByName(repository.get().name) as MavenArtifactRepository
-        logger.info("Updating URL of publishing repository '{}' to '{}'", repository.name, url)
-        repository.setUrl(url.toString())
+        project.allprojects {
+            val publishing = extensions.findByType<PublishingExtension>()
+            if (publishing != null) {
+                val repository = publishing.repositories.getByName(repository.get().name) as MavenArtifactRepository
+                logger.info("Updating URL of publishing repository '{}' to '{}'", repository.name, url)
+                repository.setUrl(url.toString())
```
vlsi, line 95: Is this safe concurrency-wise?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -40,29 +34,60 @@ import org.gradle.kotlin.dsl.withType
...
-        })
+        }
+    }
 
-        val extension = project.extensions.create<NexusPublishExtension>(NexusPublishExtension.NAME, project)
+    private fun configureTaskDependenciesForAllProjects(rootProject: Project, extension: NexusPublishExtension) {
+        rootProject.afterEvaluate {
+            rootProject.allprojects {
+                val publishingProject = this
+                pluginManager.withPlugin("maven-publish") {
```
vlsi, line 68:
```suggestion
                plugins.withId("maven-publish") {
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
vlsi: Ah, the vals can't be moved out of the loop since they reference `nexusRepo`

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -52,14 +51,13 @@ constructor(
...
-            val stagingProfileId = determineStagingProfileId(client)
-            logger.info("Creating staging repository for stagingProfileId '{}'", stagingProfileId)
-            val stagingRepositoryIdAsString = client.createStagingRepository(stagingProfileId)
-            stagingRepositoryId.invoke(stagingRepositoryIdAsString)
-            client.getStagingRepositoryUri(stagingRepositoryIdAsString)
-        }
+        val serverUrl = repository.get().nexusUrl.get()
+        val client = NexusClient(serverUrl, repository.get().username.orNull, repository.get().password.orNull, clientTimeout.orNull, connectTimeout.orNull)
+        val stagingProfileId = determineStagingProfileId(client)
+        logger.info("Creating staging repository for stagingProfileId '{}'", stagingProfileId)
```
vlsi, line 57: What do you think of adding repository name and URL to the message?
I guess staging profile alone has little value for the end-user.

```suggestion
        logger.info("Creating staging repository for {} at {}, stagingProfileId '{}'", repository.get().name, serverUrl, stagingProfileId)
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: It is, since we need to make sure the configuration for all repos and `useStaging` has been evaluated.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: Fixed.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: Done.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: It's not an invariant as it depends on `nexusRepo`. Good catch with the simpler `named<>`. 👍

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: Yes, that was done as part of #24, I've rebased onto `master` now.

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
marcphilipp: Makes sense! 👍

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -74,23 +63,31 @@ class NexusPublishPlugin : Plugin<Project> {
...
-            project.tasks.remove(project.tasks.named("initialize${capitalizedName()}StagingRepository") as Any)
-            project.tasks.remove(project.tasks.named("close${capitalizedName()}StagingRepository") as Any)
-            project.tasks.remove(project.tasks.named("release${capitalizedName()}StagingRepository") as Any)
+            project.tasks.named("initialize${capitalizedName()}StagingRepository").configure {
+                enabled = false
+            }
+            project.tasks.named("close${capitalizedName()}StagingRepository").configure {
+                enabled = false
+            }
+            project.tasks.named("release${capitalizedName()}StagingRepository").configure {
```
vlsi, line 72:
```suggestion
            project.tasks.named("initialize${capitalizedName()}StagingRepository") {
                enabled = false
            }
            project.tasks.named("close${capitalizedName()}StagingRepository") {
                enabled = false
            }
            project.tasks.named("release${capitalizedName()}StagingRepository") {
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
vlsi: `rootProject.afterEvaluate` happens after `root project` is evaluated. It might be that other projects are not evaluated yet.

vlsi: Then, you have `allprojects { plugins.withId("maven-publish") {`. It is lazy, so I see no reason to wrap `plugins.withId` with an outer `afterEvaluate` block.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -107,34 +115,32 @@ class NexusPublishPlugin : Plugin<Project> {
...
-        project.tasks.register("publishTo${nexusRepository.name.capitalize()}") {
-            description = "Publishes all Maven publications produced by this project to the '${nexusRepository.name}' Nexus repository."
-            group = PublishingPlugin.PUBLISH_TASK_GROUP
-            dependsOn(publishTasks)
-        }
-        // PublishToMavenRepository tasks may not yet have been initialized
-        project.afterEvaluate {
-            publishTasks.configureEach {
+        val mavenPublications = project.the<PublishingExtension>().publications.withType<MavenPublication>()
+        mavenPublications.all {
```
vlsi, line 125:
```suggestion
        mavenPublications.configureEach {
```

?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -107,34 +115,32 @@ class NexusPublishPlugin : Plugin<Project> {
...
-        }
-        closeTask {
-            mustRunAfter(initializeTask)
-            mustRunAfter(publishTasks)
-        }
-        releaseTask {
-            mustRunAfter(initializeTask)
-            mustRunAfter(closeTask)
-            mustRunAfter(publishTasks)
+            publishAllTask.configure {
```
vlsi, line 133:
```suggestion
            publishAllTask {
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -107,34 +115,32 @@ class NexusPublishPlugin : Plugin<Project> {
...
+                dependsOn(publishTask)
+            }
+            closeTask {
+                mustRunAfter(publishTask)
+                mustRunAfter(publishAllTask)
+            }
+            releaseTask {
+                mustRunAfter(publishTask)
+                mustRunAfter(publishAllTask)
+            }
```
vlsi, line 143: Is it required to add `closeTask { mustRunAfter(publishTask) }` (and the same for `releaseTask`)?

I guess it should be enough to add `closeTask { mustRunAfter(publishAllTask) }` just once, not multiple times (for each publication)

marcphilipp: > Is it required to add `closeTask { mustRunAfter(publishTask) }` (and the same for `releaseTask`)?

Yes, since using the `publishAllTask` is not mandatory.

> I guess it should be enough to add `closeTask { mustRunAfter(publishAllTask) }` just once, not multiple times (for each publication)

Good catch!

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
marcphilipp: Since the field is not volatile, you're probably right that this isn't 100% safe. I guess we could make sure it's set correctly using a `doFirst {}` action on every `PublishToMavenRepository` task (which is guaranteed to be executed in the same thread as the task's main action) and loading the staging repo URL from a thread-safe location (e.g. a `BuildService` with a `ConcurrentHashMap<String, URI>` for Gradle 6.1+). Other ideas?

vlsi: replacePublishingRepoUrl is called from InitializeNexusStagingRepository TaskAction which is executed in the root project.

However, it seems to update objects related to other projects which might even execute at the same time.

---

Have you tried `repository.setUrl(Provider<String>)`? Does it work?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -39,58 +37,71 @@ import org.gradle.kotlin.dsl.withType
...
+        require(project == project.rootProject) {
+            "Plugin must be applied to the root project but was applied to ${project.path}"
+        }
 
         val extension = project.extensions.create<NexusPublishExtension>(NexusPublishExtension.NAME, project)
+        configureNexusTasks(project, extension)
+        configurePublishingForAllProjects(project, extension)
+    }
 
+    private fun configureNexusTasks(project: Project, extension: NexusPublishExtension) {
```
vlsi, line 50: Have you evaluated `private fun Project.configureNexusTasks(extension: NexusPublishExtension) {` so the code looks closer to the regular `build.gradle.kts` (e.g. without duplication of `project.`)?

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
marcphilipp: I completely missed that `setUrl` takes `Object`. That should work! 👍

marcphilipp: It did work. Thanks for the idea! 🙂

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: I don't think it's worth it in this case.

marcphilipp:
> It would be good to add a simple multi-project build to the e2e suite (probably using Kotlin DSL as the single-module one uses Groovy DSL).

Done.

szpak:
I see that you have an interesting weekend :). Thanks @vlsi for your input!

It could be handled separately, but can we do something about the [warnings the new e2e project](https://github.com/gradle-nexus/publish-plugin/pull/28/checks?check_run_id=703836718#step:5:766)?

marcphilipp:
> I see that you have an interesting weekend :). Thanks @vlsi for your input!

Indeed, that was very helpful! 👍 

> It could be handled separately, but can we do something about the [warnings the new e2e project](https://github.com/gradle-nexus/publish-plugin/pull/28/checks?check_run_id=703836718#step:5:766)?

Looks like it's a known issue: https://github.com/gradle/gradle/issues/9178

PR #35, opened 2020-05-31
Log staging repo id and consumer URL when initializing the repo

### LINKED ISSUE #32: RFE: print some useful information about the Nexus staging repo and more
Print the name (and maybe the URL) of the staging repo that's been created by the plugin - maybe with `LogLevel.LIFECYCLE`.
Not sure whether there's a way to construct a URL to directly open the created staging repo in a browser (as the Nexus UI is a bit special), but that would be useful as well.

marcphilipp:
Resolves #32.

### DIFF
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
@@ -18,6 +18,7 @@ package io.github.gradlenexus.publishplugin
 
 import io.github.gradlenexus.publishplugin.internal.NexusClient
 import io.github.gradlenexus.publishplugin.internal.StagingRepositoryDescriptorRegistry
+import okhttp3.HttpUrl
 import org.gradle.api.GradleException
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -48,7 +49,10 @@ open class InitializeNexusStagingRepository @Inject constructor(
         val client = NexusClient(serverUrl, repository.username.orNull, repository.password.orNull, clientTimeout.orNull, connectTimeout.orNull)
         val stagingProfileId = determineStagingProfileId(repository, client)
         logger.info("Creating staging repository for {} at {}, stagingProfileId '{}'", repository.name, serverUrl, stagingProfileId)
-        registry.get()[repository.name] = client.createStagingRepository(stagingProfileId)
+        val descriptor = client.createStagingRepository(stagingProfileId)
+        val consumerUrl = HttpUrl.get(serverUrl)!!.newBuilder().addEncodedPathSegments("repositories/${descriptor.stagingRepositoryId}/content/").build()
+        logger.lifecycle("Created staging repository '{}' at {}", descriptor.stagingRepositoryId, consumerUrl)
```
szpak, line 54: I wonder about the logging level. With `lifecycle` it is printed in every build and it "breaks" the default "non-verbosity" promoted in the Gradle (that a task is not printed by default if it doesn't output anything). On the other hang, `info` is too verbose (and cannot be enabled per tasks) and we rather would not want to add a separate switch to turn it on and off. Therefor, I would probably leave it as is.

marcphilipp: In this case I wouldn't consider the output "noise" since the URL is useful for testing the deployed artifact.

szpak: Useful, for the projects which do that testing :-)

PR #36, opened 2020-05-31
Make staging repo description configurable

### LINKED ISSUE #31: RFE: Make the description of the Nexus staging repo configurable
It would be handy to be able to provide a custom description for the staging repo instead of just a string saying which plugin created it.
Additionally, can you mention the groupId+artifactId+version in the default description of the staging repo?

marcphilipp:
The description can now be customized on the extension or individual
tasks and now defaults to the GAV of the root project.

In addition, the User-Agent (which is shown in Nexus UI) is now
customized to `gradle-nexus-publish-plugin/<version>`.

Resolves #31.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -152,6 +152,9 @@ tasks {
         dependsOn(relocateShadowJar)
         configurations = listOf(shadowed)
         exclude("META-INF/maven/**", "META-INF/proguard/**", "META-INF/*.kotlin_module")
+        manifest {
+            attributes["Implementation-Version"] = project.version
+        }
     }
     jar {
         enabled = false
@@ -185,6 +188,9 @@ tasks {
         dependsOn(shadowJar)
         useJUnitPlatform()
         maxParallelForks = 8
+    }
+    withType<Test>().matching { it.name.startsWith("compatTest") }.configureEach {
+        systemProperty("plugin.version", project.version)
     }
     dokka {
         configuration {
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
@@ -86,8 +86,8 @@ open class NexusClient(private val baseUrl: URI, username: String?, password: St
                 ?.id
     }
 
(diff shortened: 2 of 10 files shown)


build.gradle.kts
```diff
@@ -186,6 +189,9 @@ tasks {
         useJUnitPlatform()
         maxParallelForks = 8
     }
+    withType<Test>().matching { it.name.startsWith("compatTest") }.configureEach {
+        systemProperty("plugin.version", project.version)
```
szpak, line 193: We can probably set for all the tests, not just `compatTest` (to reuse the previous section in `build.gradle`).

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
```diff
@@ -54,6 +54,12 @@ open class NexusClient(private val baseUrl: URI, username: String?, password: St
                                 .build())
                     }
         }
+        httpClientBuilder.addInterceptor { chain ->
+            val version = javaClass.`package`.implementationVersion
```
szpak, line 58: What will be used when running from IDE? Maybe some safe fallback value `undefined` or `local` or `devel` could be explicitly used?

marcphilipp: That depends how you've set up your IDE. If test execution is delegated to Gradle, it works. If IDEA is used directly, it won't.

build.gradle.kts
marcphilipp: I didn't want to add an unnecessary input to the other test tasks.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
marcphilipp: I changed it so it works either way.

szpak: Thanks. I just didn't want to get NPE or something if that property is not set.

szpak:
@marcphilipp You have some style violations.

PR #37, opened 2020-06-02
[#33] closeAndRelease* simplified tasks

### LINKED ISSUE #33: Add convenient task closeAndRelease*
To make migration from GNSP easier.

szpak:
This PR adds the following tasks:
 - `closeAndReleaseRepoNameStagingRepository` - for every Nexus repo configured
 - `closeAndReleaseStagingRepository` - for projects with just one Nexus repo configured
-  `closeAndReleaseRepository` - deprecated (legacy) task similar to above intended to make migration from GNSP easier.

Fixed #33.

I also slightly reduced verbosity of E2E tests on a CI server .

### DIFF
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
@@ -42,7 +42,6 @@ class NexusPublishPlugin : Plugin<Project> {
     companion object {
         //visibility for testing
         const val SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME = "closeAndReleaseStagingRepository"
-        const val SIMPLIFIED_LEGACY_CLOSE_AND_RELEASE_TASK_NAME = "closeAndReleaseRepository"
     }
 
     override fun apply(project: Project) {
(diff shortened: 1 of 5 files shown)


szpak:
```
Execution failed for task ':initializeSonatypeStagingRepository'.
> java.net.SocketTimeoutException: timeout
```
Twice. Retrying...

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -39,6 +39,12 @@ import org.gradle.util.GradleVersion
 @Suppress("UnstableApiUsage")
 class NexusPublishPlugin : Plugin<Project> {
 
+    companion object {
+        //visibility for testing
+        const val SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME = "closeAndReleaseStagingRepository"
+        const val SIMPLIFIED_LEGACY_CLOSE_AND_RELEASE_TASK_NAME = "closeAndReleaseRepository"
```
marcphilipp, line 45: How strongly do you feel about adding this legacy task? My gut feeling is that we don't need it.

szpak: Probably you are right. People will need to migrate to the new task name sooner or later anyway, so why postpone something what is inevitable :).

PR #48, opened 2021-02-06
Move retrying config from repo to extension and rename

### LINKED ISSUE #20: Change naming for timeout settings
> The naming is misleading.
> `RetryingConfig` + `maxRetries` means the plugin would **retry** publication multiple times, however, in practice it means it would "just wait the specified number of times while the repository is in progress".
>
>Do you think it would be better to use `timeout`-like configuration properties?
>
>For instance `connectToNexusTimeout`, `initializeRepositoryTimeout`, `singleArtifactPublishTimeout`, `closeRepositoryTimeout`, etc, etc.
>
>Note: the users do not care what the number of "status check" attempts is there. What they care is **overall** timeout value, and they might want to have something like `initial delay between status check attempts`.
>
>WDYT?

[Original comment](https://github.com/gradle-nexus/publish-plugin/pull/18#discussion_r388211811) from @vlsi in closed #18 .

I think that it's really nice idea and I don't want it forgot in closed PR.
My current timeout settings look like:
```
nexusPublishing {
    ...
    clientTimeout.set(Duration.parse("PT10M")) // 10 minutes
}
nexusStaging {
    ...
    numberOfRetries = 360 // 1 hour if 10 seconds delay
    delayBetweenRetriesInMillis = 10000 // 10 seconds
}
```

But really I just want to configure, that I'm ready to wait 1 hour for publishing staging repository.
Something like a configuration timeout for publishing artifacts to the staging repository in `nexusPublishing` block.

marcphilipp:
To avoid confusion since publishing is not retried, only checking the
repo's status when transitioning `RetryingConfig` is moved from
`NexusRepository` to `NexusPublishExtension` and renamed to
`TransitionCheckOptions`. The options specified in the extension are by
default applied to all close and release tasks but can be overridden on
each task, if need be.

Resolves #20.

### DIFF
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/AbstractTransitionNexusStagingRepositoryTask.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/AbstractTransitionNexusStagingRepositoryTask.kt
@@ -27,7 +27,6 @@ import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.TaskAction
-import org.gradle.api.tasks.options.Option
 import org.gradle.kotlin.dsl.property
 
 @Suppress("UnstableApiUsage")
@@ -43,11 +42,6 @@ abstract class AbstractTransitionNexusStagingRepositoryTask(
         set(registry.map {
             it[repository.name].stagingRepositoryId
         })
-    }
-
-    @Option(option = "staging-repository-id", description = "staging repository id to close")
-    fun setStagingRepositoryId(stagingRepositoryId: String) {
-        this.stagingRepositoryId.set(stagingRepositoryId)
     }
 
     @Internal
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/AbstractTransitionNexusStagingRepositoryTask.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/AbstractTransitionNexusStagingRepositoryTask.kt
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.gradlenexus.publishplugin
+
+import io.github.gradlenexus.publishplugin.internal.BasicActionRetrier
+import io.github.gradlenexus.publishplugin.internal.NexusClient
+import io.github.gradlenexus.publishplugin.internal.StagingRepository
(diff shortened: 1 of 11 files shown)


src/main/kotlin/io/github/gradlenexus/publishplugin/AbstractTransitionNexusStagingRepositoryTask.kt
```diff
@@ -0,0 +1,72 @@
...
+        AbstractNexusStagingRepositoryTask(objects, extension, repository) {
+
+    @Input
+    val stagingRepositoryId = objects.property<String>().apply {
+        set(registry.map {
+            it[repository.name].stagingRepositoryId
+        })
+    }
+
+    @Option(option = "staging-repository-id", description = "staging repository id to close")
```
szpak, line 44: `close` does seem to be copied from the close task. `staging repository id to transition` would be best from the domain point of view. However, the concept of repo transitioning is usually completely now known by the developers. Therefore, maybe `staging repository id to close/release` or `staging repository id to close or release`?

szpak (changes requested):

src/main/kotlin/io/github/gradlenexus/publishplugin/AbstractTransitionNexusStagingRepositoryTask.kt
marcphilipp: Good catch! How about "staging repository id to transition"?

szpak: If you think it will be understandable by users, I'm fine with it.

Alternatively, `staging repository id to transition (close or release)`.

marcphilipp: I pushed the option down to the concrete task classes now so it can be more specific.

src/main/kotlin/io/github/gradlenexus/publishplugin/TransitionCheckOptions.kt
```diff
@@ -20,9 +20,10 @@ import org.gradle.api.model.ObjectFactory
...
 import java.time.Duration
 
-open class RetryingConfig constructor(objects: ObjectFactory) {
+@Suppress("UnstableApiUsage")
+open class TransitionCheckOptions(objects: ObjectFactory) {
 
     val maxRetries = objects.property<Int>().value(60)
 
-    val delayBetween = objects.property<Duration>().value(Duration.ofSeconds(5))
+    val delayBetween = objects.property<Duration>().value(Duration.ofSeconds(10))
```
marcphilipp, line 28: @szpak Shouldn't we rather increase the number of retries?

szpak: I was thinking about that. However, a call to Sonatype Nexus every 5 seconds from all those people releasing using our plugin could potentially have a negative impact on the server.

szpak: Nevertheless, probably I exaggerate and both cases are ok (with a few seconds shorter information about successful release in the other case).

szpak (approved):

src/main/kotlin/io/github/gradlenexus/publishplugin/TransitionCheckOptions.kt
marcphilipp: Fair point, I guess waiting a few seconds longer is acceptable given the other timeouts etc. here.

PR #54, opened 2021-02-09
Make closeAndReleaseStagingRepository always available

### LINKED ISSUE #51: Consider registering createSimplifiedCloseAndReleaseTask task always
My use case: I want to hide the task from `./gradlew tasks` output since end-users would not use the task directly.

Currently, `createSimplifiedCloseAndReleaseTask` is registered in `afterEvaluate`, so I need my own `afterEvaluate` in order to hide the task (set `group=""`).

If the task was always registered, then it would be easier to configure/hide it.
An alternative option is to add an option to avoid "simplified closeAndRelease" task creation.

szpak:
Not just "after evaluate" and also when no Nexus repositories are configured. This makes it easier to hide by power users.

Also removed compatibility method for Gradle 4.x which is not supported anymore and improved task orchestration tests.

@vlsi Does it match your needs?

Fixes #51.

### DIFF
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
@@ -27,10 +27,10 @@ import org.gradle.api.publish.PublishingExtension
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.publish.maven.tasks.PublishToMavenRepository
 import org.gradle.api.publish.plugins.PublishingPlugin
-import org.gradle.api.tasks.TaskContainer
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.kotlin.dsl.create
 import org.gradle.kotlin.dsl.invoke
+import org.gradle.kotlin.dsl.named
 import org.gradle.kotlin.dsl.register
 import org.gradle.kotlin.dsl.the
 import org.gradle.kotlin.dsl.withType
@@ -55,26 +55,26 @@ class NexusPublishPlugin : Plugin<Project> {
         configurePublishingForAllProjects(project, extension, registry)
     }
 
-    private fun createRegistry(project: Project): Provider<StagingRepositoryDescriptorRegistry> {
+    private fun createRegistry(rootProject: Project): Provider<StagingRepositoryDescriptorRegistry> {
         if (GradleVersion.current() >= GradleVersion.version("6.1")) {
-            return project.gradle.sharedServices
+            return rootProject.gradle.sharedServices
                     .registerIfAbsent("stagingRepositoryUrlRegistry", StagingRepositoryDescriptorRegistryBuildService::class.java) {}
                     .map { it.registry }
         }
         val registry = StagingRepositoryDescriptorRegistry()
-        return project.provider { registry }
+        return rootProject.provider { registry }
     }
 
-    private fun configureNexusTasks(project: Project, extension: NexusPublishExtension, registry: Provider<StagingRepositoryDescriptorRegistry>) {
+    private fun configureNexusTasks(rootProject: Project, extension: NexusPublishExtension, registry: Provider<StagingRepositoryDescriptorRegistry>) {
         extension.repositories.all {
             val repository = this
-            val initializeTask = project.tasks.register<InitializeNexusStagingRepository>(
-                    "initialize${capitalizedName}StagingRepository", project.objects, extension, repository, registry)
-            val closeTask = project.tasks.register<CloseNexusStagingRepository>(
-                    "close${capitalizedName}StagingRepository", project.objects, extension, repository, registry)
-            val releaseTask = project.tasks.register<ReleaseNexusStagingRepository>(
-                    "release${capitalizedName}StagingRepository", project.objects, extension, repository, registry)
-            val closeAndReleaseTask = project.tasks.register<Task>(
+            val initializeTask = rootProject.tasks.register<InitializeNexusStagingRepository>(
+                    "initialize${capitalizedName}StagingRepository", rootProject.objects, extension, repository, registry)
+            val closeTask = rootProject.tasks.register<CloseNexusStagingRepository>(
+                    "close${capitalizedName}StagingRepository", rootProject.objects, extension, repository, registry)
+            val releaseTask = rootProject.tasks.register<ReleaseNexusStagingRepository>(
+                    "release${capitalizedName}StagingRepository", rootProject.objects, extension, repository, registry)
+            val closeAndReleaseTask = rootProject.tasks.register<Task>(
                     "closeAndRelease${capitalizedName}StagingRepository")
             closeTask {
                 description = "Closes open staging repository in '${repository.name}' Nexus instance."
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -196,26 +200,25 @@ class NexusPublishPlugin : Plugin<Project> {
...
                 val repositoryCapitalizedName = this.capitalizedName
-                val closeAndReleaseTask = rootProject.tasks.withName<Task>("closeAndRelease${repositoryCapitalizedName}StagingRepository")
+                val closeAndReleaseTask = rootProject.tasks.named<Task>("closeAndRelease${repositoryCapitalizedName}StagingRepository")
                 closeAndReleaseSimplifiedTask.configure {
                     dependsOn(closeAndReleaseTask)
                 }
             }
+        } else {
+            closeAndReleaseSimplifiedTask.configure {
+                didWork = false
```
szpak, line 220: @marcphilipp The task is reported as:
> Task :closeAndReleaseStagingRepository UP-TO-DATE

instead of SKIPPED. Is that ok or `didWork` has to be configured in some other way?

vlsi: I guess `didWork` has to be set at the runtime.

szpak: I was thinking about `doLast {}`, but that triggers some execution (instead of up-to-date).

szpak: And as I tested it is still UP-TO-DATE :-/

szpak: But maybe it works that way, anyway :)
```
> Task :closeAndReleaseStagingRepository UP-TO-DATE
Caching disabled for task ':closeAndReleaseStagingRepository' because:
  Build cache is disabled
Task ':closeAndReleaseStagingRepository' is not up-to-date because:
  Task has not declared any outputs despite executing actions.
:closeAndReleaseStagingRepository (Thread[Execution worker for ':',5,main]) completed. Took 0.001 secs.
```

szpak: Without `didWork = false`:
```
> Task :closeAndReleaseStagingRepository UP-TO-DATE
Skipping task ':closeAndReleaseStagingRepository' as it has no actions.
:closeAndReleaseStagingRepository (Thread[Execution worker for ':',5,main]) completed. Took 0.0 secs.
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -94,19 +94,23 @@ class NexusPublishPlugin : Plugin<Project> {
...
+            rootProject.tasks.named("release${capitalizedName}StagingRepository").configure {
                 enabled = false
             }
-            project.tasks.named("closeAndRelease${capitalizedName}StagingRepository").configure {
+            rootProject.tasks.named("closeAndRelease${capitalizedName}StagingRepository").configure {
                 enabled = false
             }
         }
+        rootProject.tasks.register<Task>(SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME) {
+            description = "Closes and releases open staging repository in all configured Nexus instances."
```
marcphilipp, line 111:
```suggestion
            description = "Closes and releases open staging repositories in all configured Nexus instances."
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -196,26 +200,25 @@ class NexusPublishPlugin : Plugin<Project> {
...
-
-    private fun createSimplifiedCloseAndReleaseTask(rootProject: Project, extension: NexusPublishExtension) {
+    private fun configureSimplifiedCloseAndReleaseTask(rootProject: Project, extension: NexusPublishExtension) {
+        val closeAndReleaseSimplifiedTask = rootProject.tasks.named(SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME)
         if (extension.repositories.isNotEmpty()) {
-            val closeAndReleaseSimplifiedTask = rootProject.tasks.register<Task>(SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME) {
+            closeAndReleaseSimplifiedTask.configure {
                 val repositoryNamesAsString = extension.repositories.joinToString(",") { "'${it.name}'" }
                 val instanceCardinalityAwareString = if (extension.repositories.size > 1) { "instances" } else { "instance" }
                 description = "Closes and releases open staging repository in $repositoryNamesAsString Nexus $instanceCardinalityAwareString."
```
marcphilipp, line 209:
```suggestion
                description = "Closes and releases open staging repositories in the following Nexus $instanceCardinalityAwareString: $repositoryNamesAsString"
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -196,26 +200,25 @@ class NexusPublishPlugin : Plugin<Project> {
...
-        return withType<T>().named(name)
-    }
-
-    private fun createSimplifiedCloseAndReleaseTask(rootProject: Project, extension: NexusPublishExtension) {
+    private fun configureSimplifiedCloseAndReleaseTask(rootProject: Project, extension: NexusPublishExtension) {
+        val closeAndReleaseSimplifiedTask = rootProject.tasks.named(SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME)
         if (extension.repositories.isNotEmpty()) {
-            val closeAndReleaseSimplifiedTask = rootProject.tasks.register<Task>(SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME) {
+            closeAndReleaseSimplifiedTask.configure {
                 val repositoryNamesAsString = extension.repositories.joinToString(",") { "'${it.name}'" }
```
marcphilipp, line 207:
```suggestion
                val repositoryNamesAsString = extension.repositories.joinToString(", ") { "'${it.name}'" }
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: Hmm, we could actually set `isEnabled` to `false` initially and to `true` when we configure it for at least one staging repo.

src/test/kotlin/io/github/gradlenexus/publishplugin/KotlinParameterizeTest.kt
```diff
@@ -0,0 +1,31 @@
...
+ * Change the lifecycle of a test class to [PER_CLASS][TestInstance.Lifecycle.PER_CLASS]
+ * so that parameterized tests can have non-static [MethodSource][org.junit.jupiter.params.provider.MethodSource]
+ * test argument factory methods and can be used in Kotlin tests
+ *
+ * Based on: https://blog.oio.de/2018/11/13/how-to-use-junit-5-methodsource-parameterized-tests-with-kotlin/
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class KotlinParameterizeTest
```
marcphilipp, line 31: If it works for all tests, I think we should instead configure this in `src/test/resources/junit-platform.properties` (see https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-instance-lifecycle-changing-default).

marcphilipp (approved):

src/test/kotlin/io/github/gradlenexus/publishplugin/KotlinParameterizeTest.kt
szpak: We lose a JUnit 4 "hack" for inline field initialization per test :-)

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: `isEnabled` works fine (although the output refers to `onlyIf` :) ):
```
> Task :closeAndReleaseStagingRepository SKIPPED
Skipping task ':closeAndReleaseStagingRepository' as task onlyIf is false.
:closeAndReleaseStagingRepository (Thread[Execution worker for ':',5,main]) completed. Took 0.001 secs.
```

PR #57, opened 2021-02-11
Limit JCenter to dokka and its dependencies

szpak:
Due to JCenter shut down:
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

To be merged after 1.0.0.


JavierSegoviaCordoba:
With Dokka 1.4.32, jcenter is no longer needed because kotlin.html 0.7.3 is in MavenCentral.

PR #62, opened 2021-02-18
✨ Get staging profile task

### LINKED ISSUE #60: ✨ Feature Request: GetStagingProfileTask
Is there any desired to incorporate the diagnostic `getStagingProfile` task from the `gradle-nexus-staging-plugin` to this project? I did a quick prototype and imagined it might look something like this

```kotlin
class GetStagingProfile @Inject constructor(
    objects: ObjectFactory,
    extension: NexusPublishExtension,
    repository: NexusRepository
) : AbstractNexusStagingRepositoryTask(objects, extension, repository) {

    @Optional
    @Input
    val packageGroup = objects.property<String>().apply {
        set(extension.packageGroup)
    }

    @TaskAction
    fun getStagingProfile() {
        val repository = repository.get()
        val serverUrl = repository.nexusUrl.get()
        val client = NexusClient(serverUrl, repository.username.orNull, repository.password.orNull, clientTimeout.orNull, connectTimeout.orNull)
        val stagingProfileId = client.findStagingProfileId(packageGroup = packageGroup.get()) ?: throw GradleException("Failed to find staging profile for package group: $packageGroup")
        logger.lifecycle("stagingProfileId '{}'", stagingProfileId)
    }
}
```

But I wasn't sure if this was left out unintentionally for strategic purposes - i'd be happy to contribute it with tests + update to the migration guide if this is something in line with your vision(s) for this project

As an aside, thank you for making this project! I stumbled upon it while publishing my very first artifact to maven central - this project made that a lot more approachable! ⬆️

ryandens:
Closes #60 

## Documentation Updates
I think we should add an update to the "Tasks" section of the migration guide to say something like 

```bash
./gradlew getStagingProfile
```

↓

```bash
./gradlew retrieveSonatypeStagingProfile
```

This reflects the main difference between the previous plugin project and this one: tasks are on a per `NexusRepsitory` basis, including the `RetrieveStagingProfile` task. 

## Summary of changes
First adds a new `AbstractNexusStagingRepositoryTask` subclass called `RetrieveStagingProfile`. we register a new `RetrieveStagingProfile` task for each `NexusRepository` registered in the plugin's extension. This enables users to run 
`./gradlew retrieve<NEXUS_REPOSITORY_NAME>StagingProfile` and see the staging profile ID for the repository in stdout. 

~This was initially as far as I was planning on going with this, but something @szpak mentioned in issue #60 caught my attention~
>  Or maybe there should be also an @Output field set to allow others to read it (but I'm not sure why anyone would like to consume that)

~After reading this, it struck me that we wouldn't want there to be inconsistencies between how  `GetStagingProfile` gets the staging profile and how `InitializeNexusStagingRepository` gets the staging profile. So, I went about a small refactor of the plugin to remove the responsibility of figuring out the staging profile id from `InitializeNexusStagingRepository` and instead make it a required input. Then, I configured `InitializeNexusStagingRepository` to depend on the `GetStagingProfile` which now has all the logic for getting the staging profile ID, outputting it to stdout, storing the ID in a file in the build directory, and managing the optimization of not doing the API lookup if the staging profile id is provided via configuration.~

~This not only has benefits for code re-use, but it also helps those who don't configure the stagingProfileId ever. Since we output the staging profile ID to an artifact, it can be cached and reused on subsequent builds which decreases the configuration burden on the user while still giving them access to this performance optimization.~

~I know changing `InitializeNexusStagingRepository` was not technically in the scope of what was discussed in #60 so if you would like me to remove it from the scope of this change I'd be happy to. But, it seemed to me like refactoring `InitializeNexusStagingRepository` was the responsible thing to do as I was introducing some logic duplication around how this plugin gets the staging profile id.~

Update on 2/27/21: After discussion, it was deemed that this refactor was problematic due to the conflicting nature of the different use cases of the `InitializeNexusStagingRepository` task and the diagnostic task to retrieve the staging profile

### Possible other tasks
Another thought that occurred to me was adding a task called `getStagingProfiles` or `getAllStagingProfiles` which is simply an alias for executing `get<NEXUS_REPOSITORY_NAME>StagingProfile` on each configured `NexusRepository`. I ultimately decided it was not worthwhile/might cause confusion but I'm willing to reconsider. 

### Testing gaps?
After writing up this pull request description, I realized my implementation for storing the staging repository ID was flawed, as I at first would overwrite the file storing the staging profile ID every time a new `GetStagingProfile` was run, which could lead to confusing results with multiple  `NexusRepository` destinations. The existing tests didn't catch it because there were no tests that published to two different repositories at the same time. I had to make some tweaks to the helper methods to accomplish this test, but I'm super open to feedback on how this was tested. Perhaps it would be better as a unit test than a compatTest?

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ You can configure the `connectTimeout` and `clientTimeout` properties on the `ne
 The plugin does the following:
 
 - configure a Maven artifact repository for each repository defined in the `nexusPublishing { repositories { ... } }` block in each subproject that applies the `maven-publish` plugin
+- creates a `retrieve{repository.name.capitalize()}StagingProfile` task that retrieves the staging profile id from the remote Nexus repository. This is a diagnostic task to enable setting the configuration property `stagingProfileId` in  `nexusPublishing { repositories { myRepository { ... } } }`. Specifying the configuration property rather than relying on the API call is considered a performance optimization.  
 - create a `initialize${repository.name.capitalize()}StagingRepository` task that starts a new staging repository in case the project's version does not end with `-SNAPSHOT` (customizable via the `useStaging` property) and sets the URL of the corresponding Maven artifact repository accordingly. In case of a multi-project build, all subprojects with the same `nexusUrl` will use the same staging repository.
 - make all publishing tasks for each configured repository depend on the `initialize${repository.name.capitalize()}StagingRepository` task
 - create a `publishTo${repository.name.capitalize()}` lifecycle task that depends on all publishing tasks for the corresponding Maven artifact repository
--- a/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/MethodScopeWiremockResolver.kt
+++ b/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/MethodScopeWiremockResolver.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.gradlenexus.publishplugin
+
+import com.github.tomakehurst.wiremock.WireMockServer
+import org.junit.jupiter.api.extension.AfterEachCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.junit.jupiter.api.extension.ParameterContext
+import org.junit.jupiter.api.extension.ParameterResolver
+import ru.lanwen.wiremock.ext.WiremockResolver
+import java.lang.IllegalArgumentException
+
+/**
+ * Composes a [WiremockResolver] and uses that by default. But, if a parameter is annotated
+ * with [MethodScopeWiremockResolver] it creates a new instance of the [WiremockResolver] extension and
+ * manages its lifecycle in the scope of that [ExtensionContext]
+ */
+class MethodScopeWiremockResolver(
+    private val inner: WiremockResolver = WiremockResolver()
+) : ParameterResolver by inner, AfterEachCallback {
+
+    /**
+     * Checks to see if a method the parameter is annotated with [MethodScopedWiremockServer]. If it is, we first verify
+     */
+    override fun resolveParameter(parameterContext: ParameterContext, extensionContext: ExtensionContext): Any {
+        return if (parameterContext.isAnnotated(MethodScopedWiremockServer::class.java)) {
+            if (!parameterContext.parameter.type.isAssignableFrom(WireMockServer::class.java)) {
+                throw IllegalArgumentException("Annotated type must be a WireMockServer")
+            }
+            return getStore(extensionContext)
+                .getOrComputeIfAbsent(Keys.LOCAL_RESOLVER, { WiremockResolver() }, WiremockResolver::class.java)
+                .resolveParameter(parameterContext, extensionContext)
+        } else {
+            inner.resolveParameter(parameterContext, extensionContext)
+        }
+    }
+
+    override fun afterEach(context: ExtensionContext) {
+        getStore(context).get(Keys.LOCAL_RESOLVER, WiremockResolver::class.java)?.afterEach(context)
+        inner.afterEach(context)
+    }
+
+    /**
+     * helper method for get getting a [ExtensionContext.Store] specific to a test method
+     */
+    private fun getStore(context: ExtensionContext): ExtensionContext.Store {
+        return context.getStore(ExtensionContext.Namespace.create(javaClass))
+    }
+
+    /**
+     * Keys for storing and accessing [MethodScopeWiremockResolver]s
+     */
+    enum class Keys {
+        LOCAL_RESOLVER
+    }
+
+    /**
+     * Decorates a parameter also annotated with [WiremockResolver.Wiremock]
+     */
+    @Target(allowedTargets = [AnnotationTarget.VALUE_PARAMETER])
+    @Retention(AnnotationRetention.RUNTIME)
+    annotation class MethodScopedWiremockServer
+}
--- a/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
+++ b/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
@@ -48,13 +48,12 @@ import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.io.TempDir
-import ru.lanwen.wiremock.ext.WiremockResolver
 import ru.lanwen.wiremock.ext.WiremockResolver.Wiremock
 import java.nio.file.Files
 import java.nio.file.Path
 
 @Suppress("FunctionName") // TODO: How to suppress "kotlin:S100" from SonarLint?
-@ExtendWith(WiremockResolver::class)
+@ExtendWith(MethodScopeWiremockResolver::class)
 class NexusPublishPluginTests {
 
     companion object {
@@ -233,9 +232,7 @@ class NexusPublishPluginTests {
     }
 
     @Test
-    fun `publishes to two Nexus repositories`() {
-        val otherServer = WireMockServer()
-        otherServer.start()
+    fun `publishes to two Nexus repositories`(@MethodScopeWiremockResolver.MethodScopedWiremockServer @Wiremock otherServer: WireMockServer) {
         projectDir.resolve("settings.gradle").write("""
             rootProject.name = 'sample'
         """)
--- a/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
+++ b/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
@@ -111,6 +111,32 @@ class NexusPublishPluginTests {
     }
 
     @Test
+    fun `can get StagingProfileId from Nexus`() {
+        writeDefaultSingleProjectConfiguration()
+        //and
+        buildGradle.append("""
+            nexusPublishing {
+                repositories {
+                    sonatype {
+                        nexusUrl = uri('${server.baseUrl()}')
+                        allowInsecureProtocol = true
+                        //No staging profile defined
+                    }
+                }
+            }
+        """)
+        //and
+        stubGetStagingProfilesForOneProfileIdGivenId(STAGING_PROFILE_ID)
+
+        val result = run("retrieveSonatypeStagingProfile")
+
+        assertSuccess(result, ":retrieveSonatypeStagingProfile")
+        assertThat(result.output).containsOnlyOnce("Received staging profile id: '$STAGING_PROFILE_ID' for package org.example")
+        // and
+        assertGetStagingProfile(1)
+    }
+
+    @Test
     fun `publish task depends on correct tasks`() {
         projectDir.resolve("settings.gradle").write("""
             rootProject.name = 'sample'
@@ -204,6 +230,73 @@ class NexusPublishPluginTests {
                 .withRequestBody(matchingJsonPath("\$.data[?(@.description == 'org.example:sample:0.0.1')]")))
         assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.pom")
         assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.jar")
+    }
+
+    @Test
+    fun `publishes to two Nexus repositories`() {
+        val otherServer = WireMockServer()
+        otherServer.start()
+        projectDir.resolve("settings.gradle").write("""
+            rootProject.name = 'sample'
+        """)
+        projectDir.resolve("build.gradle").write("""
+            plugins {
+                id('java-library')
+                id('maven-publish')
+                id('io.github.gradle-nexus.publish-plugin')
+            }
+            group = 'org.example'
+            version = '0.0.1'
+            publishing {
+                publications {
+                    mavenJava(MavenPublication) {
+                        from(components.java)
+                    }
+                }
+            }
+            nexusPublishing {
+                repositories {
+                    myNexus {
+                        nexusUrl = uri('${server.baseUrl()}')
+                        snapshotRepositoryUrl = uri('${server.baseUrl()}/snapshots/')
+                        allowInsecureProtocol = true
+                        username = 'username'
+                        password = 'password'
+                    }
+                    someOtherNexus {
+                        nexusUrl = uri('${otherServer.baseUrl()}')
+                        snapshotRepositoryUrl = uri('${otherServer.baseUrl()}/snapshots/')
+                        allowInsecureProtocol = true
+                        username = 'someUsername'
+                        password = 'somePassword'
+                    }
+                }
+            }
+        """)
+
+        val otherStagingProfileId = "otherStagingProfileId"
+        val otherStagingRepositoryId = "orgexample-43"
+        stubStagingProfileRequest("/staging/profiles", mapOf("id" to STAGING_PROFILE_ID, "name" to "org.example"))
+        stubStagingProfileRequest("/staging/profiles", mapOf("id" to otherStagingProfileId, "name" to "org.example"), wireMockServer = otherServer)
+        stubCreateStagingRepoRequest("/staging/profiles/$STAGING_PROFILE_ID/start", STAGED_REPOSITORY_ID)
+        stubCreateStagingRepoRequest("/staging/profiles/$otherStagingProfileId/start", otherStagingRepositoryId, wireMockServer = otherServer)
+        expectArtifactUploads("/staging/deployByRepositoryId/$STAGED_REPOSITORY_ID")
+        expectArtifactUploads("/staging/deployByRepositoryId/$otherStagingRepositoryId", wireMockServer = otherServer)
+
+        val result = run("publishToMyNexus", "publishToSomeOtherNexus")
+
+        assertSuccess(result, ":initializeMyNexusStagingRepository")
+        assertSuccess(result, ":initializeSomeOtherNexusStagingRepository")
+        assertThat(result.output).containsOnlyOnce("Created staging repository '$STAGED_REPOSITORY_ID' at ${server.baseUrl()}/repositories/$STAGED_REPOSITORY_ID/content/")
+        assertThat(result.output).containsOnlyOnce("Created staging repository '$otherStagingRepositoryId' at ${otherServer.baseUrl()}/repositories/$otherStagingRepositoryId/content/")
+        server.verify(postRequestedFor(urlEqualTo("/staging/profiles/$STAGING_PROFILE_ID/start"))
+            .withRequestBody(matchingJsonPath("\$.data[?(@.description == 'org.example:sample:0.0.1')]")))
+        otherServer.verify(postRequestedFor(urlEqualTo("/staging/profiles/$otherStagingProfileId/start"))
+            .withRequestBody(matchingJsonPath("\$.data[?(@.description == 'org.example:sample:0.0.1')]")))
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.pom")
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.jar")
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.pom", stagingRepositoryId = otherStagingRepositoryId, wireMockServer = otherServer)
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.jar", stagingRepositoryId = otherStagingRepositoryId, wireMockServer = otherServer)
     }
 
     @Test
@@ -352,6 +445,7 @@ class NexusPublishPluginTests {
 
         val result = gradleRunner("initializeMyNexusStagingRepository").buildAndFail()
 
+        // we assert that the first task that sends an HTTP request to server fails as expected
         assertOutcome(result, ":initializeMyNexusStagingRepository", FAILED)
         assertThat(result.output).contains("SocketTimeoutException")
     }
@@ -732,14 +826,18 @@ class NexusPublishPluginTests {
     }
 
     @SafeVarargs
-    private fun stubStagingProfileRequest(url: String, vararg stagingProfiles: Map<String, String>) {
-        server.stubFor(get(urlEqualTo(url))
+    private fun stubStagingProfileRequest(
+        url: String,
+        vararg stagingProfiles: Map<String, String>,
+        wireMockServer: WireMockServer = server
+    ) {
+        wireMockServer.stubFor(get(urlEqualTo(url))
                 .withHeader("User-Agent", matching("gradle-nexus-publish-plugin/.*"))
                 .willReturn(aResponse().withBody(gson.toJson(mapOf("data" to listOf(*stagingProfiles))))))
     }
 
-    private fun stubCreateStagingRepoRequest(url: String, stagedRepositoryId: String) {
-        server.stubFor(post(urlEqualTo(url))
+    private fun stubCreateStagingRepoRequest(url: String, stagedRepositoryId: String, wireMockServer: WireMockServer = server) {
+        wireMockServer.stubFor(post(urlEqualTo(url))
                 .willReturn(aResponse().withBody(gson.toJson(mapOf("data" to mapOf("stagedRepositoryId" to stagedRepositoryId))))))
     }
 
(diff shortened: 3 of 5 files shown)


src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -206,6 +206,102 @@ class NexusPublishPluginTests {
...
+        otherServer.verify(postRequestedFor(urlEqualTo("/staging/profiles/$otherStagingProfileId/start"))
+            .withRequestBody(matchingJsonPath("\$.data[?(@.description == 'org.example:sample:0.0.1')]")))
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.pom")
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.jar")
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.pom", stagingRepositoryId = otherStagingRepositoryId, wireMockServer = otherServer)
+        assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.jar", stagingRepositoryId = otherStagingRepositoryId, wireMockServer = otherServer)
+    }
+
+    @Test
+    fun `can get StagingProfileId from Nexus`() {
```
szpak, line 280: We don't enforce a declared order of tests, nor skill other tests after the first failure, but for logic order in operations performed in read, I would put it as the second test in the class (right after `must be applied to root project`).

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
```diff
@@ -0,0 +1,74 @@
...
+                clientTimeout.orNull,
+                connectTimeout.orNull
+            )
+            val packageGroup = packageGroup.get()
+            client.findStagingProfileId(packageGroup) ?: throw GradleException("Failed to find staging profile for package group: $packageGroup")
+        } else {
+            configuredStagingProfileId
+        }
+
+        logger.lifecycle("Received staging profile id: '{}' for package org.example", stagingProfileId)
```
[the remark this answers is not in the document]
szpak: I like the idea of reusing the `getStagingProfile` task by the initialization task. However, I see a side effect which disturbs the silence of a regular Gradle tasks executions. On every publishing there will be a few lines displayed (task name, empty line and the message). Especially, having `stagingProfileId` explicitly configured it would be unexpected.

Possibly, we could detect if the task was explicitly called from the command line (e.g. using `gradle.startParameter.taskNames`) and only then display that message at the lifecycle level (otherwise on info). That might be not very elegant, but maybe @marcphilipp knows some Gradle trick to tweak it.

In addition, `Received staging profile id` is not fully true for the explicitly configured value (it was not received) and might be confusing for people migrating from GNSP.  Maybe there could be two variants of the output:
> Configured staging profile id: ...

and 
> Received staging profile id: ...

used depending on configuration. Nevertheless, I'm not fixed on that.

szpak (changes requested):
Really informative PR description, thanks!

I went through the code and in general I like it. Chaining tasks makes sense, but there is one side effect related to logging. I put some remarks inline.

I also requested [e2e tests](https://github.com/gradle-nexus/publish-plugin/actions/runs/579625776) on your PR.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -206,6 +206,102 @@ class NexusPublishPluginTests {
         assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.jar")
     }
 
+    @Test
+    fun `publishes to two Nexus repositories`() {
+        val otherServer = WireMockServer()
+        otherServer.start()
```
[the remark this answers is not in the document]
ryandens: Ah, I definitely meant to come back here and clean this up prior to opening the PR, but I totally forgot! I tried to leverage the JUnit extension for providing a new instance of a WireMockServer, but it seems to not be supported. I think we have a couple of options
1. Decorate the WireMockServer with a `Closeable` implementation that calls the appropriate shutdown methods on the server
2. Write a small junit extension to manage the lifecycle of another WireMockServer. This is nice because it lets us re-use some of the existing components of the junit5 extension already in use by this project.



This is what the `Closeable` implementation/usgae would look like


```

    class CloseableWireMockServerContainer(val port: Int, private val server: WireMockServer = WireMockServer(port)) : Closeable,
        Supplier<WireMockServer> {
        override fun close() {
            server.stop()
        }

        override fun get(): WireMockServer {
            return server
        }
    }

@Test
    fun `publishes to two Nexus repositories`() {
      CloseableWireMockServerContainer(0). use {
      
      val otherServer = it.get()
      // omitted, rest of test

  }
}
```

Or, this is what the JUnit extension would look like 
```
/**
 * Composes a [WiremockResolver] and uses that by default. But, if a parameter is annotated
 * with [MethodScopeWiremockResolver] it creates a new instance of the [WiremockResolver] extension and
 * manages its lifecycle in the scope of that [ExtensionContext]
 */
class MethodScopeWiremockResolver(
    private val inner: WiremockResolver = WiremockResolver(),
    private val localResolvers: MutableMap<String, WiremockResolver> = mutableMapOf()
) : ParameterResolver by inner, AfterEachCallback {

    /**
     * Checks to see if a method the parameter is annotated with [MethodScopedWiremockServer]. If it is, we first verify
     */
    override fun resolveParameter(parameterContext: ParameterContext, extensionContext: ExtensionContext): Any {
        return if (parameterContext.isAnnotated(MethodScopedWiremockServer::class.java)) {
            if (!parameterContext.parameter.type.isAssignableFrom(WireMockServer::class.java)) {
                throw IllegalArgumentException("Annotated type must be a WireMockServer")
            }
            return localResolvers.getOrPut(extensionContext.uniqueId, { WiremockResolver() })
                .resolveParameter(parameterContext, extensionContext)
        } else {
            inner.resolveParameter(parameterContext, extensionContext)
        }
    }

    override fun afterEach(context: ExtensionContext) {
        val localWiremockResolver = localResolvers[context.uniqueId]
        if (localWiremockResolver != null) {
            localWiremockResolver.afterEach(context)
            localResolvers.remove(context.uniqueId)
        }
        inner.afterEach(context)
    }

    /**
     * Decorates a parameter also annotated with [WiremockResolver.Wiremock]
     */
    @Target(allowedTargets = [AnnotationTarget.VALUE_PARAMETER])
    @Retention(AnnotationRetention.RUNTIME)
    annotation class MethodScopedWiremockServer
}

  @Test
    fun `publishes to two Nexus repositories`(@MethodScopedWiremockServer @Wiremock otherServer: WireMockServer) {
    // omitted, rest of test
```

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
```diff
@@ -0,0 +1,74 @@
...
+     * This property is [Optional] because it is only required if [NexusRepository.stagingProfileId] is not provided
+     */
+    @Input
+    @Optional
+    val packageGroup = objects.property<String>().apply {
+        set(extension.packageGroup)
+    }
+
+    @OutputFile
+    val stagingProfileIdOutput: RegularFileProperty = objects.fileProperty()
```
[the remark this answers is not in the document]
ryandens: I'm unfamiliar with the `@Output` annotation - could you point me to some documentation? I don't see it in the `org.gradle.api.tasks` package and was under the impression that task outputs generally were associated with a file/directory

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
ryandens: Agreed! When I first wrote this test, I didn't realize I would be adding this into the chain of execution of the publish tasks 😃

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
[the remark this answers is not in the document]
ryandens: This makes sense, I agree that we don't want this to be printed out every time this task is executed. My first thought was to add a configuration property in the `GetStagingProfile` which determines whether or not the diagnostic data is printed out. That task property could be configured in the `NexusPublishPlugin`. However, I don't see a way of doing that without creating a new registration of the task with a different name to allow for the two different configurations to exist

In my opinion, that means the `GetStagingProfile` task is doing too much, as it trying to be a diagnostic tool and a first-class step in the publishing task chain. I think it would make sense to register a  `diagnosticGet${repository.name.capitalize()}StagingProfile` task for each repository that depends on the `get${repository.name.capitalize()}StagingProfile` task. The `GetStagingProfile` class is solely responsible for determining the staging profile id, either from configuration or from the HTTP API of the configured nexus repository. the `DiagnosticGetStagingProfile` consumes the staging profile ID and is instead focused on printing out the right message, depending on whether or not a `NexusRepository.stagingProfileId` is configured. This means that the `initialize` task can rely strictly upon the `GetStagingProfile` task but users still have access to a diagnostic message via the CLI if they deliberately execute the task.  This might look like:

```kotlin
/**
 * Diagnostic task for printing out the provided [stagingProfileId] in an informative message. The message
 * format differs based on whether or not the [NexusRepository.stagingProfileId] was configured explicitly
 * or not. If it was not, the [stagingProfileId] is presumed to have been obtained from a remote source,
 * like the [NexusRepository] HTTP API.
 */
@Suppress("UnstableApiUsage")
@Incubating
open class DiagnosticGetStagingProfile @Inject constructor(
    objects: ObjectFactory,
    extension: NexusPublishExtension,
    repository: NexusRepository
) : AbstractNexusStagingRepositoryTask(objects, extension, repository) {


    /**
     * This property is [Optional] because it is only required if [NexusRepository.stagingProfileId] is not provided
     */
    @Input
    @Optional
    val packageGroup = objects.property<String>().apply {
        set(extension.packageGroup)
    }

    @Input
    val stagingProfileId = project.objects.property(String::class)

    @TaskAction
    fun diagnosticGetStagingProfile() {
        val repository = repository.get()
        val configuredStagingProfileId = repository.stagingProfileId.orNull
        if (configuredStagingProfileId == null) {
            logger.lifecycle("Received staging profile id: '{}' for package '{}' for repository '{}'", stagingProfileId.get(), packageGroup.get(), repository.capitalizedName)
        } else {
            logger.lifecycle("Configured staging profile id: '{}' for repository '{}'", configuredStagingProfileId, repository.capitalizedName)
        }
    }
}
```

In `NexusPublishPlugin`, we would register the task and configure the properties and task dependencies for each repository

```kotlin
val diagnosticGetStagingProfile = rootProject.tasks.register<DiagnosticGetStagingProfile>("diagnosticGet${repository.name.capitalize()}StagingProfile")

getStagingProfileTask {
    description = "Gets a staging profile id for a given repository. If the staging profile id is configured, this task simply retrieves it from the extension that configured it. If the it is not, the staging profile id is retrieved from the NexusRepository via its HTTP API a packageGroup must be configured on the project that applied this plugin."
      stagingProfileIdOutput.set(project.layout.buildDirectory.file("publish-plugin/get${capitalizedName}StagingProfileOutput.txt"))
}
diagnosticGetStagingProfile {
    description = "Displays staging profile id for a given repository. This is a diagnostic task to get the value and put it into the NexusRepository configuration closure as stagingProfileId."
    stagingProfileId.set(getStagingProfileTask.map { it.stagingProfileIdOutput.get().asFile.readText() })
}

```

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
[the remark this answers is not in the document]
vlsi: 1. `@Output` does not exist
2. By default, `Gradle` would consider the task `UP-TO-DATE` if all inputs and outputs are the same.

It looks like `OutputFile` is OK here.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -76,6 +77,14 @@ class NexusPublishPlugin : Plugin<Project> {
                     "release${capitalizedName}StagingRepository", rootProject.objects, extension, repository, registry)
             val closeAndReleaseTask = rootProject.tasks.register<Task>(
                     "closeAndRelease${capitalizedName}StagingRepository")
+            getStagingProfileTask {
+                description = "Gets and displays a staging profile id for a given repository and package group. This is a diagnostic task to get the value and put it into the NexusRepository configuration closure as stagingProfileId."
+                group = PublishingPlugin.PUBLISH_TASK_GROUP
```
vlsi, line 82: I'm not sure it is worth displaying the task in the common output for `./gradlew tasks`.
Don't get me wrong, but I believe very few end-users would call `./gradlew get... StagingProfile`.

Please keep `group=null` so the task does not appear in `gradlew tasks` output

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -176,6 +190,9 @@ class NexusPublishPlugin : Plugin<Project> {
         val mavenPublications = project.the<PublishingExtension>().publications.withType<MavenPublication>()
         mavenPublications.configureEach {
             val publication = this
+            initializeTask {
+                dependsOn(getStagingProfileTask)
+            }
```
vlsi, line 195: In practice, this dependency should be automatically set by Gradle when you configure `stagingProfileId.set(getStagingProfileTask.map...`

I guess it does not hurt, however, it is not needed either.

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        val configuredStagingProfileId = repository.stagingProfileId.orNull
+        val stagingProfileId: String = if (configuredStagingProfileId == null) {
+            val client = NexusClient(
+                repository.nexusUrl.get(),
+                repository.username.orNull,
+                repository.password.orNull,
+                clientTimeout.orNull,
+                connectTimeout.orNull
+            )
```
vlsi, line 67: I guess this should be factored into an `internal val` since the client instantiation is duplicated in all the tasks.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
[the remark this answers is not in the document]
marcphilipp: > Does JUnit Jupiter has an equivalent of `cleanup:` in Spock - to be always executed at the end of that particular test only?

It does not but it would be fairly easy to write an extension that collects `Closable` instances.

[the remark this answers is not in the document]
marcphilipp: > Does JUnit Jupiter has an equivalent of `cleanup:` in Spock - to be always executed at the end of that particular test only?

It does not but it would be fairly easy to write an extension that collects `Closable` instances.

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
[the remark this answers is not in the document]
marcphilipp: I don't think this task should ever be up-to-date since it depends on state on the server. Thus, if we remove the `@OutputFile` property and add an `@Internal val stagingProfileId: Property<String>` we don't need to write a file.

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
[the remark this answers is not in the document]
marcphilipp: I think we should at most log this on `info` level.

README.md
```diff
@@ -120,6 +120,7 @@ You can configure the `connectTimeout` and `clientTimeout` properties on the `ne
 The plugin does the following:
 
 - configure a Maven artifact repository for each repository defined in the `nexusPublishing { repositories { ... } }` block in each subproject that applies the `maven-publish` plugin
+- creates a `get${repository.name.capitalize()}StagingProfile` task that either retrieves the staging profile id from the remote Nexus repository or from the configuration property `stagingProfileId` in  `nexusPublishing { repositories { myRepository { ... } } }`. Specifying the configuration property rather than relying on the API call is considered a performance optimization.
```
marcphilipp, line 123: I think we should name the task in a way that makes it clearer that it makes a call to a server, e.g. `retrieve${repository.name.capitalize()}StagingProfileId` or `find${repository.name.capitalize()}StagingProfileId`.

marcphilipp (changes requested):

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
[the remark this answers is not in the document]
szpak: Sorry @ryandens for twisting the things around. I liked the idea with dependent tasks, however, the recent comments by Marc helped me realize that fetching the staging repository id from a server for diagnostic purpose is not the same as resolving staging repository id for the initialize tasks (and others). In the first case, server should be always asked (should ignore configured stagingProfileId), output should be printed on lifecycle level (people always want to see it) and it should be always not up-to-date. On the other side, for the initialize task is perfectly fine to get pre-configured value (if available), be quiet and play nicely with caching (assuming that `stagingProfileId` was set).

As a result, I think that we need `Retrieve<<Repo>>StagingProfile` task to implement the first case (always using  `client.findStagingProfileId()`). The second one could be covered as in the original version in `InitializeNexusStagingRepository`. As it is recommended to set `stagingProfileId` in the project configuration (for the performance reasons), I don't really see practical use case where someone would need to get that value from server on its own to reuse (with `stagingRepositoryId` the situation is completely different). WDYT?

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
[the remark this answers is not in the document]
szpak: With my [last comment](https://github.com/gradle-nexus/publish-plugin/pull/62#discussion_r579826757) the whole case with different logging would disappear.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: Assuming its one-time per project (or even a group of projects) role, I agree that keeping it mentioned only in the documentation (in README) would be enough.

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
szpak: I believe, I proposed it in one of my earlier comments, but suggested to move it to a separate PR, after this one is merged to do not mix the things.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
marcphilipp: Good point. We don't actually want that dependency when `stagingProfileId` is set explicitly. Thus, we should remove it.

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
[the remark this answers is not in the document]
szpak: @ryandens Are you ok with the proposed modifications (simplifications)?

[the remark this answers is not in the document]
ryandens: Yes, this seems reasonable! I'll take a stab at getting this all sorted out

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
ryandens: Definitely happy to defer to you all on this one, but for what it's worth the first thing I did when trying to figure out how to get the staging profile with this new plugin was running `./gradlew tasks` and checking the output for a similar task

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -206,6 +231,71 @@ class NexusPublishPluginTests {
         assertUploadedToStagingRepo("/org/example/sample/0.0.1/sample-0.0.1.jar")
     }
 
+    @Test
+    fun `publishes to two Nexus repositories`(@MethodScopeWiremockResolver.MethodScopedWiremockServer @Wiremock otherServer: WireMockServer) {
```
ryandens, line 235: this test isn't strictly necessary in order for us to feel confident about the changes that are now being made in this PR, but if you find it valuable I'm happy to leave it in.

szpak: Yes, please.

szpak (approved):

src/main/kotlin/io/github/gradlenexus/publishplugin/RetrieveStagingProfile.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.TaskAction
+import org.gradle.kotlin.dsl.property
+import javax.inject.Inject
+
+/**
+ * Diagnostic task for retrieving the [NexusRepository.stagingProfileId] for the [packageGroup] from the provided [NexusRepository] and logging it
+ */
+@Incubating
```
marcphilipp, line 31: @szpak Not sure if we should use this annotation. WDYT?

marcphilipp (commented):
Why don't we use the new task to retrieve the staging profile id for the initialize task?

src/main/kotlin/io/github/gradlenexus/publishplugin/RetrieveStagingProfile.kt
szpak: Personally, I like to use `@Incubating` for the new tasks/features to gather feedback and possibly have a clear way to fix spotted issues even in a non-backward compatible way. However, you don't like the "beta" state of that feature or just a Gradle annotation?

szpak:
> Why don't we use the new task to retrieve the staging profile id for the initialize task?

@marcphilipp I asked the PR's author about that a comment earlier - see my [reasoning](https://github.com/gradle-nexus/publish-plugin/pull/62#discussion_r579826757). It simplifies the general construction and we eventually still use the common logic from the client.

src/main/kotlin/io/github/gradlenexus/publishplugin/RetrieveStagingProfile.kt
marcphilipp: Ok, fine with me but we should have some common understanding when to do this. IIUC you'd like to do this for all new features. Is that correct?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/MethodScopeWiremockResolver.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.lang.IllegalArgumentException
+
+/**
+ * Composes a [WiremockResolver] and uses that by default. But, if a parameter is annotated
+ * with [MethodScopeWiremockResolver] it creates a new instance of the [WiremockResolver] extension and
+ * manages its lifecycle in the scope of that [ExtensionContext]
+ */
+class MethodScopeWiremockResolver(
+    private val inner: WiremockResolver = WiremockResolver(),
+    private val localResolvers: MutableMap<String, WiremockResolver> = mutableMapOf()
```
marcphilipp, line 34: Extensions should not have mutable state but store it in the `ExtensionContext.Store` but we/I can address that after this PR is merged.

marcphilipp (approved):

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/MethodScopeWiremockResolver.kt
ryandens: Ah, definitely agreed! This was easy enough for me to fix now 😃  - I'm a fan of Extension API so always enjoy the opportunity to work with it more

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/MethodScopeWiremockResolver.kt
```diff
@@ -0,0 +1,77 @@
...
+    override fun afterEach(context: ExtensionContext) {
+        getStore(context).get(Keys.LOCAL_RESOLVER, WiremockResolver::class.java)?.afterEach(context)
+        inner.afterEach(context)
+    }
+
+    /**
+     * helper method for get getting a [ExtensionContext.Store] specific to a test method
+     */
+    private fun getStore(context: ExtensionContext): ExtensionContext.Store {
+        return context.getStore(ExtensionContext.Namespace.create(javaClass, context.requiredTestMethod))
```
marcphilipp, line 61: There's no need to add the test method to the namespace since the extension context is already scoped to the method level.

ryandens: Thanks! I couldn't remember the rules around this so I went looking in the user guide and derived it from this example: https://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-before-after-execution  - but fixed!

marcphilipp: Good point! That example should indeed be simplified.

src/main/kotlin/io/github/gradlenexus/publishplugin/RetrieveStagingProfile.kt
szpak: It would be safe, but maybe it would be to defensive, as we - in general - do not expose API to others (at least, it is not the main goal of that extension). How does it look like in JUnit? Are you satisfied with the strategy adopted there?

src/main/kotlin/io/github/gradlenexus/publishplugin/GetStagingProfile.kt
szpak: Created #69.

szpak:
Thanks @ryandens for your PR (and patience ;-) ). It will be a part of 1.1.0.

ryandens:
Awesome, thanks for accepting my contribution and working through the feedback with me! Hoping this makes it easier for other folks to adopt this plugin 😄

PR #64, opened 2021-02-22
Clarified staging repository isolation in README

### LINKED ISSUE #63: Explain and document the staging repository isolation
Nice plugin! (again!)

How does this plugin handle parallel publications from *different* GitHub projects that share the *same* group id, and publish to the *same* nexus repo (nexus oss)? Let's call it "staging repository isolation" use case.

Let's make it a little more concrete. In Mockito we have "org.mockito:mockito-kotlin" and "org.mockito:mockito-scala". Those are separate GH projects with independent release cycles. We have a strong preference to keep the same group "org.mockito" (branding, compatibility). What happens if both of those projects use the publish-plugin, and trigger release at the same time, to the same nexus repo (OSS sonatype)? Specifically, will the project use separate staging repositories?

Ask: please document this use case :) ping @szpak

mockitoguy:
Fixes #63


szpak:
Thanks @mockitoguy!

PR #107, opened 2021-06-07
Update dependencies & improve build performance

nhajratw:
Just some minor updates.

FYI, upgrading gradle to 7.0.2 (the latest), caused test failures that I didn't know how to solve, so I just updated to the latest 6.x series.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,3 @@
+org.gradle.caching=true
+org.gradle.parallel=true
+org.gradle.vfs.watch=true
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=7faa7198769f872826c8ef4f1450f839ec27f0b4d5d1e51bade63667cbccd205
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,11 +6,11 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 plugins {
     `kotlin-dsl`
     `maven-publish`
-    id("com.gradle.plugin-publish") version "0.14.0"
-    id("com.diffplug.spotless") version "5.12.1"
+    id("com.gradle.plugin-publish") version "0.15.0"
+    id("com.diffplug.spotless") version "5.12.5"
     id("com.github.johnrengelman.shadow") version "6.1.0"
     id("org.jetbrains.gradle.plugin.idea-ext")
-    id("com.github.ben-manes.versions") version "0.38.0"
+    id("com.github.ben-manes.versions") version "0.39.0"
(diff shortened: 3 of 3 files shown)


marcphilipp (approved):

marcphilipp:
@nhajratw Thanks! 👍

gradle.properties
```diff
@@ -0,0 +1,3 @@
+org.gradle.caching=true
+org.gradle.parallel=true
+org.gradle.vfs.watch=true
```
szpak, line 3: I'm not convinced if we should enable it at the project level. It's rather user environment specific decision.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=7faa7198769f872826c8ef4f1450f839ec27f0b4d5d1e51bade63667cbccd205
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
```
szpak, line 3: Why the is `distributionSha256Sum` missing?

szpak (changes requested):

gradle.properties
marcphilipp: It's enabled by default in Gradle 7 so I think that's ok.

gradle/wrapper/gradle-wrapper.properties
marcphilipp: The IDEA issue with it has been fixed in the meantime. I'll re-add it when upgrading to Gradle 7.

nhajratw: I just ran `./gradlew wrapper --gradle-version=6.9` -- this was the result.

marcphilipp: @szpak was referring to the additional `--gradle-distribution-sha256-sum` option (cf. https://docs.gradle.org/current/userguide/gradle_wrapper.html#configuring_checksum_verification)

nhajratw: yes, i know -- I'm just saying that the wrapper command seems to have removed that line. Not sure if it's not required....

Anyway , per that link you posted, it seems like you could add a GitHub Action to validate the wrapper checksum. Might be a good idea.

szpak: @nhajratw The following command is better:
```
./gradlew wrapper --gradle-version 7.0.2 --distribution-type bin --gradle-distribution-sha256-sum 0e46229820205440b48a5501122002842b82886e76af35f0f3a069243dca4b3c
```
(with checksum available [here](https://gradle.org/release-checksums/))

szpak: > Anyway , per that link you posted, it seems like you could add a GitHub Action to validate the wrapper checksum. Might be a good idea.

We have one, so maybe it even did that :)
https://github.com/gradle-nexus/publish-plugin/actions/workflows/gradle-wrapper-validation.yml

szpak: > @szpak was referring to the additional --gradle-distribution-sha256-sum option

I missed that your command Marc. Thanks.

marcphilipp: This action validates the checksum of the wrapper JAR that is checked in in `gradle/wrapper`. The `distributionSha256Sum` is about validating the checksum of the Gradle distribution that is downloaded by the wrapper.

PR #124, opened 2021-11-12
Remove Dokka

marcphilipp:
No longer generate Javadoc using Dokka since it keeps causing issues and
we only have a single production class with Javadoc anyway.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -11,7 +11,6 @@ plugins {
     id("com.github.johnrengelman.shadow") version "6.1.0"
     id("org.jetbrains.gradle.plugin.idea-ext")
     id("com.github.ben-manes.versions") version "0.39.0"
-    id("org.jetbrains.dokka") version "1.4.32"
     id("org.ajoberstar.stutter") version "0.6.0"
 }
 
@@ -94,7 +93,6 @@ dependencies {
 }
 
 java {
-    withJavadocJar()
     withSourcesJar()
 }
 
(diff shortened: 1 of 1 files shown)


szpak (approved):

szpak:
I created https://github.com/gradle-nexus/publish-plugin/issues/125 to track further improvements in that topic.

vlsi:
Removing javadoc is unfortunate as it blocks publishing the plugin to OSSRH.
AFAIK, GradlePluginPortal does not support PGP signing for the plugins (see https://github.com/gradle/gradle/issues/10182), so I tend to publish all my Gradle plugins to both OSSRH and GPP.

build.gradle.kts
```diff
@@ -94,7 +93,6 @@ dependencies {
 }
 
 java {
-    withJavadocJar()
     withSourcesJar()
```
szpak, line 96:
```suggestion
    withJavadocJar()
    withSourcesJar()
```

As @vlsi pointed our, it could be problematic to have `-javadoc` jar completely missing. Restoring that one line generates "empty" java file, which should be enough for Maven Central requirements (once we will decide to publish also there in the future).

szpak:
> Removing javadoc is unfortunate as it blocks publishing the plugin to OSSRH.

You are right. As we currently release only to the Plugin Portal, there was an idea in the past to publish also to Maven Central. Luckily, there is a quick workaround to just generate an "empty" JAR with JavaDocs. See my suggestion above.

build.gradle.kts
vlsi: I agree, publishing an empty javadoc jar would be reasonable workaround

PR #126, opened 2021-11-14
provide default URIs for different sonatype hosts

tynn:
The default URIs should be easily configureable for the original and for the s01 host. Therefore I added an enum to provide the base urls and resolve the nexus and snapshot repository URLs.


src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
```diff
@@ -21,22 +21,15 @@ import org.gradle.api.Action
...
 
 @Suppress("UnstableApiUsage")
 internal class DefaultNexusRepositoryContainer(delegate: NamedDomainObjectContainer<NexusRepository>) : NexusRepositoryContainer, NamedDomainObjectContainer<NexusRepository> by delegate {
 
     override fun sonatype(): NexusRepository = sonatype {}
 
-    override fun sonatype(closure: Closure<*>): NexusRepository {
-        return sonatype(ConfigureUtil.configureUsing(closure))
-    }
+    override fun sonatype(closure: Closure<*>): NexusRepository = sonatype(ConfigureUtil.configureUsing(closure))
```
vlsi, line 30: Closure overloads are not needed, and they can be removed

tynn: Why are these not needed any more?

Anyhow I'd prefer to keep this out of the scope of this PR: It would change the `NexusRepositoryContainer` interface as well and should be done with caution.

vlsi: Methods with `Action<...>` work for ages for both Groovy DSL and Kotlin DSL, so there's no need to duplicate methods, and if only `Action<...>` methods are present, then there's less likely there will be something wrong with ConfigureUtil.

---

> Anyhow I'd prefer to keep this out of the scope of this PR

I see you touch Closure-related line, so I would suggest just removing it instead of trying to maintain it.

tynn: I acutally prefer a `Closure` over an `Action` when using the _Groovy_ DSL 😅

vlsi: @tynn ,
1) Gradle makes sure Action works for Groovy just fine
2) `Closure<*>` provides **no** clue on the receiver type. In other words, `Closure<*>` is like Java 1.4 with `Object`, `Vector`, `Hashtable` , and so on.
3) Lots of Gradle developers recommend stopping using `Closure` in plugin apis. For instance: https://twitter.com/CedricChampeau/status/1349669642646806531

vlsi: See https://github.com/gradle/gradle/blob/732148799d2e7b0efe07b2649d5b80b4b3ab7c3f/subprojects/model-core/src/main/java/org/gradle/util/ConfigureUtil.java#L37

vlsi: I filed https://github.com/gradle/gradle/issues/18973 so Gradle issues a warning in case developers use Closure or T.() for the API.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepository.kt
```diff
@@ -27,11 +27,23 @@ import javax.inject.Inject
...
     @Input
-    val snapshotRepositoryUrl = project.objects.property<URI>()
+    val nexusUrl = project.objects.property<URI>().apply {
+        set(nexusHost.map { it.resolve("/service/local/") })
+    }
+
+    @Input
+    val snapshotRepositoryUrl = project.objects.property<URI>().apply {
+        set(nexusHost.map { it.resolve("/content/repositories/snapshots/") })
+    }
```
marcphilipp, line 46: I think having three properties is weird because it would allow configuring a host that's not even used in `nexusUrl` and `snapshotRepositoryUrl`.

README.md
```diff
@@ -37,14 +37,13 @@ nexusPublishing {
...
 
 ```gradle
 nexusPublishing {
     repositories {
-        sonatype {  //only for users registered in Sonatype after 24 Feb 2021
-            nexusUrl.set(uri("https://s01.oss.sonatype.org/service/local/"))
-            snapshotRepositoryUrl.set(uri("https://s01.oss.sonatype.org/content/repositories/snapshots/"))
+        sonatype {
+            nexusHost = s01
         }
```
marcphilipp, line 47: How about the following?

```suggestion
        sonatype("s01.oss.sonatype.org")
```

and

```suggestion
        sonatype("s01.oss.sonatype.org") {
			...
		}
```

src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
marcphilipp: This is required due to https://github.com/gradle/gradle/issues/9987

README.md
vlsi: `sonatype("s01.oss.sonatype.org")` is not really auto-completable :-/

I guess the sole point of `sonatype` is to have a pre-defined configuration for OSSRH, so having enums looks reasonable.

---

`sonatype("nexus.acme.com")` would look a bit weird

marcphilipp: I don't really feel comfortable hard-coding server names like `s01`. Has Sonatype shared any public roadmap for that server? Will there be `s02`, `s03` etc. at some point?

vlsi: So far there are only two instances.
Of course, it might sense to have a generic API that receives hostname, however, I believe, it is way more important and user-friendly to have an enum or something for `s01`.

Frankly speaking, I do not understand how `s01..` allows OSSRH to resolve the performance issue, and it is really really surprising for me they went with user-facing DNS change.

vlsi: Yet another alternative is to probe (e.g. list profiles request or something) the server and auto-detect the host among `oss.` and `s01.` if user omits the host.

src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
vlsi: It looks like `publish-plugin` uses `newInstance` explicitly: https://github.com/gradle-nexus/publish-plugin/blob/0aea95597e2eeaf9275a286254894bad29b3d1e1/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt#L53-L55, so `#9987` should not be important

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepository.kt
tynn: `nexusUrl` would be the default only. It wouldn't apply if you override any of the URLs directly.

README.md
tynn: Eventually I'd like to get rid of the manual declaration of the default (which differs for users).

Sonatype suggested, that they might spin up another host when they start having issues with `s01`.

So for now there's only `s01.oss` and `oss`. I had implementations using strings before, but that could result in user error on configuration, so I choose to use an enum instead. Also the enum is extensible when new hosts spin up.

Also this approach works nicely in _Groovy_ and _Kotlin_.

tynn: A possible shorthand/replacement could be 

    sonatype('s01')

Still relying on an `internal enum class NexusHost` for verification.

`toEnum(NexusHost::class.java, nexusHost)` works nicely.

marcphilipp: I still think configuring the hostname explicitly is clearer instead of just `s01`.

@szpak WDYT?

szpak: Passing 's01' (or enum) might be somehow confusing, however, on the other hand, we originally created `sonatype()` to make the configuration straightforward for a default scenario, so there could be an option to use it with  `s01` in some similar convenient way.

Having just two Sonatype servers, we could have just `sonatype()` and `sonatypeS01()`. It will not scale, but I don't know about Sonatype plans to add another machines (assuming that `oss` has a lot of more projects that `s01` it might take some time to reach its capacity). Alternatively, I think that we could treat `sonatype('s01')` not as passing a host name, but rather a server "name", with some enum under the hood, allowing to control passed value and help to determine real server name (NOT just string concatenation "s01" + ".oss.sonatype..." accepting everything). Having `s01` easily completable in Idea (will it work also in build.gradle in Groovy?) is a plus for me.

szpak: > Frankly speaking, I do not understand how s01.. allows OSSRH to resolve the performance issue, and it is really really surprising for me they went with user-facing DNS change.

@vlsi It's just my speculation, but I think that Sonatype doesn't have a problem with just serviing the released artifacts (which could be and is delegated to some CDN), but rather with the release process (creating a staging repository, closing it and releasing/dropping). It is handled by Nexus Pro 2 which might not be able to handle XXX simultaneous operations (even on a very large machine(s)). Having it scattered with a new independent system - for new people/projects - at least does not increase the original workload with new projects added every single day, again and again.

marcphilipp: I don't think we should be more concerned about expressiveness in the build script than auto completion. 

> Having just two Sonatype servers, we could have just `sonatype()` and `sonatypeS01()`.

I would be okay with that or `sonatype("s01")`. @szpak Which variant would you prefer?

tynn: I added a commit implementing `sonatype(s01)`...

README.md
```diff
@@ -42,9 +42,7 @@ nexusPublishing {
 ```gradle
 nexusPublishing {
     repositories {
-        sonatype {
-            nexusHost = s01
-        }
+        sonatypes(s01)
```
vlsi, line 45:
```suggestion
        sonatype(s01)
```

README.md
```diff
@@ -42,9 +42,7 @@ nexusPublishing {
 ```gradle
 nexusPublishing {
     repositories {
-        sonatype {
-            nexusHost = s01
-        }
+        sonatype(s01)
```
szpak, line 45: Doesn't it require any import? That's a reason, I am reluctant in using enum directly. I remember situations in Groovy build.gradle where Idea shows red error on an import (even thought, it was correctly resolved by Gradle itself). Has it improved over time?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -541,6 +541,64 @@ class NexusPublishPluginTests {
         assertThat(result.output).contains("SocketTimeoutException")
     }
 
+    @Test
+    fun `uses s01 default URLs for sonatype repos in Groovy DSL`() {
```
szpak, line 545: These two tests are very similar to the two for the old instance. Wouldn't be possible to cover them with `junit-jupiter-params`?

README.md
tynn: `s01` is a member of the `NexusRepositoryContainer` interface

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
tynn: Done :+1:

szpak (approved):

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepository.kt
```diff
@@ -27,11 +27,18 @@ import javax.inject.Inject
 @Suppress("UnstableApiUsage")
 open class NexusRepository @Inject constructor(@Input val name: String, project: Project) {
 
+    @get:Internal
+    internal lateinit var nexusHost: NexusHost
+
```
marcphilipp, line 32: I think this should be a constructor parameter instead.

marcphilipp (changes requested):

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepository.kt
tynn: @marcphilipp me too. How would I provide it then? I might have missed that API, but for me it looks like it's not possible to provide the argument when `create` is called.

marcphilipp: By passing it here like `name` and `project`:
https://github.com/gradle-nexus/publish-plugin/blob/1c7b5099ac276e7a157ea4dfb60b67ee2c2da538/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt#L55

vlsi: I'm afraid Gradle provides no way to add extra arguments when creating containers with `project.container` API: https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#container-java.lang.Class-org.gradle.api.NamedDomainObjectFactory-

The factory is hard-coded to have a single name argument only: https://docs.gradle.org/current/javadoc/org/gradle/api/NamedDomainObjectFactory.html

tynn: Yes, I noticed the same. I was expecting an API like `newInstance`, but couldn't find it. So I thought I might have missed some dark magic in _Gradle_...

marcphilipp: You're right, I overlooked that little detail... 😬 

Could you please remove the `var` and add a method that configures the urls based on a `NexusHost` parameter?

vlsi: What's wrong with var?
It allows users to modify host after object creation which might be useful

tynn: @marcphilipp I'm not sure I understand your intent.

In my opinion, `lateinit var nexusHost: NexusHost` would be the most honest approach. Also it fits better into the lazy approach of properties. Evenutally `nexusHost` will only be evaluated when necessary.

marcphilipp: My intent is to avoid having potentially contradictory properties in `NexusRepository`.

vlsi: The only "contradictory" case I could imagine is when user modifies `nexusUrl` and `snapshotRepositoryUrl` properties.
Do you really mean you do not like that?

If that is the case, the only way to avoid this ambiguity is to make different classes for `OssrhNexusHost` (that is driven by `nexusHost`, and which provides no way to modify `nexusUrl`; and `CustomNexusHost` that lacks `nexusHost` field, and that allows `nexusUrl` customization.
However, I think it would be overkill, and it might be too restrictive for the users

vlsi: However, it might be sligtly better to use `project.objects.property` for `nexusHost`, which would allow clients to configure the host lazily (e.g. based on the value of a Gradle property).

marcphilipp: > Do you really mean you do not like that?

Yes, I really mean that.

> If that is the case, the only way to avoid this ambiguity is to make different classes for `OssrhNexusHost` (that is driven by `nexusHost`, and which provides no way to modify `nexusUrl`; and `CustomNexusHost` that lacks `nexusHost` field, and that allows `nexusUrl` customization.

We cannot do that without breaking backwards compatibility.

tynn: @marcphilipp all done :+1:

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepositoryContainer.kt
```diff
@@ -19,9 +19,17 @@ package io.github.gradlenexus.publishplugin
 import groovy.lang.Closure
 import org.gradle.api.Action
 import org.gradle.api.NamedDomainObjectContainer
+import org.gradle.internal.Actions
```
vlsi, line 22: Does it really make sense to depend on internal class here? I would prefer {} instead

tynn: Done :+1:

marcphilipp (approved):

marcphilipp:
@szpak Do you want to take another look?

szpak:
> @szpak Do you want to take another look?

No, thank you. You know much better all the tricks & tips with the Gradle API and Kotlin :-). If the PR looks satisfactorily good for you, please merge it.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusHost.kt
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.gradlenexus.publishplugin
+
+import java.net.URI
+
+enum class NexusHost(private val baseUrl: String) {
```
vlsi, line 21: `NexusHost` is a too wide naming. `Nexus` can be hosted by anyone, so I think `NexusHost` is a wrong name.

What do you think of `OssrhHost` or `SonatypeOssrhHost`?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepositoryContainer.kt
```diff
@@ -21,7 +21,14 @@ import org.gradle.api.Action
 import org.gradle.api.NamedDomainObjectContainer
 
 interface NexusRepositoryContainer : NamedDomainObjectContainer<NexusRepository> {
-    fun sonatype(): NexusRepository
-    fun sonatype(closure: Closure<*>): NexusRepository
-    fun sonatype(action: Action<in NexusRepository>): NexusRepository
+    @Suppress("unused")
+    val s01 get() = NexusHost.S01
```
vlsi, line 25: Should there be `oss()` property as well? It does not sound fair to have one accessor only :)

szpak: The old server it used by default, so for consistency maybe there could be `oss()`, but in the majority of cases is would not be used. In the remaining cases - some conditional behavior configured in code in some other plugin, probably it should be possible to refer just to the enum.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusHost.kt
marcphilipp: Good call! I'd prefer `SonatypeOssrhHost`.

marcphilipp: On the other hand, this could be used for other, non-Sonatype OSSRH hosts in the future. 🤔

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusRepositoryContainer.kt
marcphilipp: I think we should add `oss` for people who want to be explicit, e.g. when having to use different hosts in different projects.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusHost.kt
marcphilipp, line 21: Let's make this a regular data class and just have `oss` and `s01` constants in `NexusRepositoryContainer` so it's open for extension.

marcphilipp (changes requested):

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusHost.kt
tynn: I'd make this an `fun interface` or normal `class` instead. There's no need for any of the data class feature overhead.

tynn: Since we're going to make it configureable for the user, I'll keep the name as `NexusHost`.

vlsi: Then it makes sense to add a method like `nexus(name: String, urlPrefix: String)` to `NexusRepositoryContainer`.
It would enable users to create repositories (e.g. non-sonatype ones) with their own URLs. On the positive side, they won't have to manually override both `releaseUrl` and `snapshotUrl`. WDYT?

tynn: While we're at it. Should we just provide a string with the `baseUrl` instead? @marcphilipp @vlsi

vlsi: `sonatype(host: EnumLikeType, ...)` should better have enum-like parameter since currently we know only two well-known hosts: `oss.sonatype....` and `s01....`

On the other hand, a separate method might be helpful for creating nexus server with `String`-based `baseUrl` so users can add their own servers.

`sonatype("https://nexus.internal.example.com")` would look weird in the build script. What would `sonatype` even mean there?

tynn: Now I do get your point regarding the `nexus` function. I think we're getting off track here. `NexusHost` would be a good naming if we want to configure any nexus. Here we only want to configure the `SonatypeNexusHost` or `SonatypeHost`, so maybe we should keep the enum as is. Maybe deprecate the implicit `sonatype` factory and add the suggested `nexus` factory as a semantic differentiation.

    sonatype(oss) { ... }
    sonatype(s01) { ... }

    nexus("name") { ... }

Do you know if Nexus is fully configurable or does it provide static paths?

tynn:
@marcphilipp I gave it another thought, but mostly kept it as it was before. Mostly I changed the semantics to be perfectly applicable.

I renamed `NexusHost` to `SonatypeOSSRepositoryHost`. Thus it's more specific and as a sealed class still enum like.

As a next step I dropped the `s10` variable and replaced it with `oss` holding the `s01` instead. Thus it can be like the reverse host name itself.

Like this it's possible to use the `sonatype` functions like:

    sonatype { ... } // configures URLs for oss.sonatype.org
    sonatype(oss) { ... } // configures URLs for oss.sonatype.org
    sonatype(oss.s01) { ... } // configures URLs for s01.oss.sonatype.org

The `sonatype` therefore serves it's original purpose as a preset for sonatype with full semantics for the different hosts and could be used as a normal container entry (the first example, just that urls are preset as well).

I'd propose to not change anything else there and keep it strictly for sonatype only.

As a next step it might be useful to have a `nexusHost` input property and resolve a default `nexusUrl` and `snapshotRepositoryUrl` from it. But it should not be part of a factory function like `sonatype`, even if it's convenient:

    myNexus {
        nexusHost.set('https://nexus.internal.example.com')
    }

Issue #19, opened 2020-03-18
Ability to create and release staging repository in separate calls (e.g. after manual verification)


szpak:
<hr/>

### Proposal by @vlsi placed originally [here](https://github.com/gradle-nexus/publish-plugin/issues/4#issuecomment-553595035), extracted to a separate (this) issue.

<hr/>

I propose the following workflow:

## Stage release artifacts

1) As release manager starts the release, they decide release version. For instance, it could be `v1.0-rc1`. The version could be configured in the sources or passed via parameters (it does not matter)
2) `initializeStagingRepository` takes that version number, and puts it to the staging repository metadata. I'm not sure what are the possibilities, however, I'm sure the version can be put to the "comment" field
3) `initializeStagingRepository` produces an `@Output` with `stagingRepositoryId`. It could save the id to file (files are commonly expected task inputs and outputs). It could provide the id via `Property<String>`.

## Verify release
4) Everybody might want to verify the artifacts. They know "release candidate index", and they can pass it to the build script, so it locates the relevant staging repository, dowloads the artifacts, and validates them.

## Publish release
The release manager remembers "release candidate version" to be released, and it passes that parameter to the build script.
4) `findStagingRepository(version=...)` task queries Nexus, and it detects the staging repository id. Then regular `closeRepositoryOrWhatever` task could be used to close&promote the repository.

^^^ Note: the above is not something that must be used by everybody. It is just a sample.
Of course, one could combine all the tasks into a single Gradle execution.

In other words, I suggest using Nexus Server for storage of "repository id" rather than "repository extension".

<hr/>

### The subsequent [comment](https://github.com/gradle-nexus/publish-plugin/issues/4#issuecomment-569282962) by @visi:

<hr/>

In 99.42% of the cases, users want reasonable comments for their staging repositories.
For instance, I have the following:
<img width="997" alt="Nexus staging repositories" src="https://user-images.githubusercontent.com/213894/71520762-c0b9d680-28ce-11ea-88da-f8dd766b12de.png">

It corresponds to ~5 or so "release candidates".
The version tree between them looks as follows:
<img width="558" alt="git version tree" src="https://user-images.githubusercontent.com/213894/71520792-f068de80-28ce-11ea-9c24-8d89fe7e2d6f.png">

I have multiple "closed" repositories, and I can easily test each version individually, and later I can decide which one is better.

What if I want to promote `v3.0-rc5` as the final release?
[...]

PR #145, opened 2022-06-30
Add FindStagingRepository task to find staging repository by its description

vlsi:
The task would be helpful when "publish staged repository" is implemented as a separate step.

See #19


szpak:
> @szpak , @marcphilipp , any thoughts?

I'm "out of office" and I will not be able to review that before August :-/.

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
```diff
@@ -37,6 +37,7 @@ class NexusPublishE2ETests : BaseGradleTest() {
         //when
         val result = run(
             "publishToSonatype",
+            "findSonatypeStagingRepository",
```
vlsi, line 40: I could probably move that to a different test if you think it is needed, however, I would love to hear some feedback first

vlsi:
@szpak , @marcphilipp , do you think you could have some cycles for reviewing the change?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -885,6 +885,24 @@ class NexusPublishPluginTests {
...
+    fun `should find staging repository by description`() {
+        //given
+        writeDefaultSingleProjectConfiguration()
+        writeMockedSonatypeNexusPublishingConfiguration()
+        //and
+        val stagingRepository = StagingRepository(STAGED_REPOSITORY_ID, StagingRepository.State.OPEN, false)
+        val responseBody = getStagingReposWithOneStagingRepoWithGivenIdJsonResponseAsString(stagingRepository)
+        stubGetStagingReposForStagingProfileIdWithResponseStatusCodeAndResponseBody(STAGING_PROFILE_ID, 200, responseBody)
+
+        val result = run("findSonatypeStagingRepository")
```
szpak, line 898: It fails. I suspect there is something with the regex.

>  No staging repository found for stagingProfileId: someProfileId, descriptionRegex: \Qorg.example:sample:0.0.1\E(\s|$) (found 1 in total)

is returned, even if Wiremock returns:
>               "description": "org.example:sample:0.0.1",

In general, I think that integration test should just the main scenario - the repo is found (and maybe a failure message). How would you see testing the regex itself (and surrounding logic) in more "unit" way?

src/main/kotlin/io/github/gradlenexus/publishplugin/FindStagingRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+    fun findStagingRepository() {
+        val repository = repository.get()
+        val serverUrl = repository.nexusUrl.get()
+        val client = NexusClient(serverUrl, repository.username.orNull, repository.password.orNull, clientTimeout.orNull, connectTimeout.orNull)
+        val stagingProfileId = determineStagingProfileId(repository, client)
+        logger.info("Fetching staging repositories for {} at {}, stagingProfileId '{}'", repository.name, serverUrl, stagingProfileId)
+        val descriptionRegex = descriptionRegex.get()
+        val descriptor = client.findStagingRepository(stagingProfileId, Regex(descriptionRegex))
+        logger.lifecycle("Staging repository for {} at {}, stagingProfileId '{}', descriptionRegex '{}' is '{}'", repository.name, serverUrl, stagingProfileId, descriptionRegex, stagingRepositoryId)
+        this.stagingRepositoryId.set(stagingRepositoryId)
```
szpak, line 68: It seems, there is something wrong here. The output is:
```
> Task :findSonatypeStagingRepository
Staging repository for sonatype at http://localhost:35705, stagingProfileId 'someProfileId', descriptionRegex '\Qorg.example:sample:0.0.1\E(\s|$)' is task ':findSonatypeStagingRepository' property 'stagingRepositoryId'
```

And in fact the logged uses the input which itself is set in the next line (however, `this.stagingRepositoryId` and `stagingRepositoryId` seems to point to the same object/property).

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
```diff
@@ -91,6 +91,17 @@ open class NexusClient(private val baseUrl: URI, username: String?, password: St
             ?.id
     }
 
+    fun findStagingRepository(stagingProfileId: String, descriptionRegex: Regex): StagingRepositoryDescriptor {
+        val response = api.getStagingRepositories(stagingProfileId).execute()
+        if (!response.isSuccessful) {
+            throw failure("find staging repository, stagingProfileId: $stagingProfileId, descriptionRegex: $descriptionRegex", response)
+        }
+        val totalNumberOfFoundRepos = response.body()?.data?.count()    //TODO: Consider printing other repo IDs and descriptions on INFO if none matched
```
szpak, line 99: To consider.

szpak (changes requested):

szpak:
Sorry for delay with reviewing :-/. I gave it a try today and planned to finish it and merge. However, after I wrote mocked integration tests and I've found some things to discuss (commented inline). Of course, if you are still interested in getting it done.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -72,6 +72,12 @@ class NexusPublishPlugin : Plugin<Project> {
             val initializeTask = rootProject.tasks.register<InitializeNexusStagingRepository>(
                 "initialize${capitalizedName}StagingRepository", rootProject.objects, extension, repository, registry
             )
+            val findStagingRepository = rootProject.tasks.register<FindStagingRepository>(
+                "find${capitalizedName}StagingRepository", rootProject.objects, extension, repository, registry
+            )
+            findStagingRepository {
+                description = "Finds the staging repository for ${repository.name}"
```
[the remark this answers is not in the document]
vlsi: For now I incline to keep the task hidden, as it is probably not very useful on its own

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
vlsi: The way to test regex is to factor `firstOrNull` to a separate method, or even add `internal` method like `fun findStagingRepository(response: Response<Dto<List<ReadStagingRepository>>>, regex: Regex)`.
Then it could be tested without executing a call.

However, I do not see that to be a pattern yet: other tests are integration-only.

vlsi: The bug was `\b` instead of `\\b` 🤷

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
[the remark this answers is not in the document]
vlsi: See https://github.com/gradle-nexus/publish-plugin/pull/145#discussion_r960557725

Technically speaking, the end-to-end could be staging and releasing with two different executions.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
szpak: Things happen. Regexes definitely can be tricky, so - even more - it  would be good to have it thoroughly tested. Gradle doesn't help ease unit testing, but `NexusClient` is not so Gradle bound. Maybe we could the second constructor in `NexusClient` accepting `NexusApi`? Then, in tests we could just inject there a `NexusApi` mock and stub `api.getStagingRepositories()`. WDYT?

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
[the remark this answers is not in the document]
szpak: > Technically speaking, the end-to-end could be staging and releasing with two different executions.

In fact those e2e tests are rather "smoke tests". I wanted to have it verified that the basic scenario(s) pass(es). Therefore, writing about some verification, I rather had in mind checking that task was effectively executed (not failed, so it found matching repository) and the "heuristic" (to find just one staging repository) was not used. WDYT?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
vlsi: It might make sense to rework the approach to testing, however, it sounds like out of scope for the PR.

I would say a more important test here is to test stage and publish in independent gradle executions.

However, it was not clear if you would ever review/accept the feature, so I did not spent time on testing

vlsi:
We should probably define what happens if multiple repositories match.

We might add an enum like: FAIL (default), USE_LATEST

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
szpak: > It might make sense to rework the approach to testing, however, it sounds like out of scope for the PR.

The basic case with matching passes, so we could just create a separate issue to rework it and write more unit tests for that regexp (just to have it covered, assuming changes in the future).

> I would say a more important test here is to test stage and publish in independent gradle executions.

Good point. The parameter can be already passed from the command line in the second invocation, so it *should* behave here in the same way, however, definitely it would be good to test it :). I would prefer to do it with Wiremock (yes, I know that `NexusPublishPluginTests` is asking to be refactored, but let's sort the other things out first :-/).

> However, it was not clear if you would ever review/accept the feature, so I did not spent time on testing

AFAIR, you requested that feature to be used (among others) in the Apache Foundation projects where "staged" artifacts should be additionally tested/verified. That a valid use case, so - in general - I had been always open to accept it (especially that you provide the implementation :-) ). I just didn't precise which "August" I had in mind ;-).

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
```diff
@@ -91,6 +91,17 @@ open class NexusClient(private val baseUrl: URI, username: String?, password: St
             ?.id
     }
 
+    fun findStagingRepository(stagingProfileId: String, descriptionRegex: Regex): StagingRepositoryDescriptor {
+        val response = api.getStagingRepositories(stagingProfileId).execute()
+        if (!response.isSuccessful) {
+            throw failure("find staging repository, stagingProfileId: $stagingProfileId, descriptionRegex: $descriptionRegex", response)
+        }
+        val totalNumberOfFoundRepos = response.body()?.data?.count()
+        return response.body()?.data?.firstOrNull { it.description?.contains(descriptionRegex) == true }
```
szpak, line 100: > We should probably define what happens if multiple repositories match.
> 
> We might add an enum like: FAIL (default), USE_LATEST

FAIL is much safer and I agree should be a default behavior. What are the cases for `USE_LATEST`? Some previous execution errors with leftover staging repositories? I wonder, if a developer should be forced to log in to the Nexus UI to clean it up, or you believe that USE_LATEST should be "safe" (if explicitly chosen)?

Btw, I still assume that without explicitly calling the `find*StagingRepository` task, the staging repository id should be called manually from the command line (if init and close/release are not performed in the same execution), right?

(move to the inline command to make it easier to follow)

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
vlsi: I've created a separate "two stage" release testin E2ETests

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
vlsi: I've added a test to verify cases when the repository is not found

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/NexusClient.kt
vlsi: I've hard-coded "fail if multiple repositories found"

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
```diff
@@ -47,4 +47,40 @@ class NexusPublishE2ETests : BaseGradleTest() {
             assertSuccess(":releaseSonatypeStagingRepository")
         }
     }
+
+    @ParameterizedTest(name = "{0}")
+    @ValueSource(strings = ["nexus-publish-e2e-minimal", "nexus-publish-e2e-multi-project"])
+    fun `release project to real Sonatype Nexus in two executions`(projectName: String) {
```
szpak, line 53: 1. It will most likely fails, as the version is [timestamp based](https://github.com/gradle-nexus/nexus-publish-e2e-minimal/blob/6a7fc8d7895b0f3aea803bf7037cef05bfa1d438/build.gradle#L9-L11) to avoid problem described in https://github.com/gradle-nexus/publish-plugin/pull/153 (in addition the version is not properly set in the multi-module e2e project as I mentioned in the comment a few days ago).
2. I can run it from a side branch to verify (if you think `1.` is wrong or once there is a workaround - what do you propose?), but in general, I'm still not sure, if we should call it on every merge/push to master, as e2e tests (despite their name) are rather smoke/sanity tests for critical scenarios.

szpak (changes requested):

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -889,6 +889,52 @@ class NexusPublishPluginTests {
         )
     }
 
+    @Test
+    fun `should not find staging repository by wrong description`() {
+        // given
+        writeDefaultSingleProjectConfiguration()
+        buildGradle.append("version='2.3.4-so staging repository is not found'")
```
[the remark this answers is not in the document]
vlsi: Could we do that when the value will actually be reused?

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
vlsi: >as the version is timestamp based to avoid problem described

What do you think of passing the version on the command line instead?

>I'm still not sure, if we should call it on every merge/push to master, as e2e tests (despite their name) are rather smoke/sanity tests for critical scenarios.

There are only 2 merges to master per year. Is adding a test really a problem?
If you absolutely want to reduce the number of interactions with Central, you might consider launching a random subset of tests every time.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -889,6 +889,52 @@ class NexusPublishPluginTests {
         )
     }
 
+    @Test
+    fun `should not find staging repository by wrong description`() {
```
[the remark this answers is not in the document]
vlsi: I probably forgot to copy the test before editing it.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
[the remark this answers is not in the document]
vlsi: If you mean to reuse the version only withing the method, I incline the current code is easier to maintain. I copy-pasted the actual output to the assertion message, and that is enough to update the assertion.

Note: the assertion message will likely need updates if the message is reworded. I incline it might happen, and every time it would require re-inserting the actual "non-matching" version

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
szpak: > There are only 2 merges to master per year. Is adding a test really a problem?

It depends of a time window. I see 17 commits to master in 2 weeks :). With 2x30+ projects releases released to Nexus.

In general, this is not a big problem right now (especially that I plan to reduce number of e2e tests execution #181 to mainly master and cron), but I'm really happy being able to perform the real e2e testing with the real Sonatype Nexus and it would be problematic to get banned (or something). E2E test should cover the main/critical scenarios and from my experience, especially in the "enterprise stuff", they tend to grow (with all the maintenance issues). I'm really curious, how many project will use the new mode (and publishing performed in two phases).

Anyway, let's see how many build it trigger in the coming weeks. In the worst case, we could add a switch to just call them (the new one) from cron, not for every master build.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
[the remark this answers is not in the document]
szpak: I meant `2.3.4-so staging repository is not found` which is already reused could a `val` and "... ${x} ..." in the assertion message. I could be more precise. WDYT?

Btw, I have mixed feeling about that kind of big exact matching. Maybe checking just for 4-5 different element would be enough? Nevertheless, I'm not convinced, if my proposal would better in that case, so keep it as is and let's see how easy/hard it would be to maintain/read in practice.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -76,6 +76,16 @@ class NexusPublishPlugin : Plugin<Project> {
                 repository,
                 registry
             )
+            val findStagingRepository = rootProject.tasks.register<FindStagingRepository>(
```
szpak, line 79: Btw, we need to document it somewhere, to let people know. Would you see it in README or we should just mention it in the README, but provide sample used on the wiki?

src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt
szpak: > What do you think of passing the version on the command line instead?

Yes, give me a moment, I'm playing with some idea.

szpak: Would adding `-PoverriddenVersion=somethingUnique` twice with the same (unique) value generated in `given` be enough?

https://github.com/gradle-nexus/nexus-publish-e2e-minimal/commit/6a6fbdb3b34a8cacc54b634cf75031050f738a74
https://github.com/gradle-nexus/nexus-publish-e2e-multi-project/commit/a2de98f251529bee6e567f742828febd34f99db9

szpak: Btw, how would you approach the issue with:

> :nexus-publish-e2e-multi-project:unspecified

reported for the root project? Here, in the multi-module Kotlin project, the things are slightly complicated as the version is set in `buildSrc/library-publishing-conventions.gradle.kts` just for submodules?

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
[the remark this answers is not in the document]
vlsi: A bigger string gives much more information to the maintainer on the exact thing that was expected.
If you add an assert like  <<`2.3.4-so staging repository is not found` should be present in the output>>, then it would be unclear where and why the substring should appear.

vlsi:
After having some more thought,  I incline to a different design: what if `InitializeNexusStagingRepository` was like "find an existing staging repository or create a new one if missing"?

Then it would handle cases like `./gradlew publishToSonatype; ./gradlew closeSonatypeStagingRepository; ./gradlew releaseSonatypeStagingRepository` automatically.

Well, it might worth adding an extra property to `InitializeNexusStagingRepository` to specify the behavior:

* `FIND_OR_CREATE` -- reuses an existing staging repository if present, or creates a new one
* `CREATE_NEW` -- always creates a new staging repository
* `FIND_OR_FAIL` -- searches for an existing repository, and fails if the repo is missing
* `FIND_OR_NULL` -- searches for an existing repository, and just keeps `null` value if the repo is missing

WDYT?

----

>:nexus-publish-e2e-multi-project:unspecified

Well, apparently, the staging repository is created after the repository description of the root project. So there are two approaches here:
a) allow overriding the version of the root project (it is the only version that is really needed, and `find..` task does not use subproject versions)
b) allow overriding `repositoryDescription` property in test execution

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -42,16 +44,83 @@ open class InitializeNexusStagingRepository @Inject constructor(
...
+                                    serverUrl,
+                                    stagingProfileId,
+                                    descriptionRegex
+                                )
+                                null
+                            }
+                        }
+                    }
+                }
+            }
```
szpak, line 114: Sorry, but that is too complex as a "one-liner" assigning a value to `descriptor`. 5 levels of nested statement and clumsy conditional logic. It would need to be refactored into smaller pieces, however, please take a look at the "general comment" to new approach before doing any actions here.

szpak (changes requested):

szpak:
In general, changing the behavior of the init task (and effectively also `publishToSonatype`) is not a backward compatible change. I incline to postpone it to the next release to avoid confusion from people upgrading "just to fix deprecation warning".

Talking about the feature itself, I might imagine a situation when, someone calls `publishToSonatype` and `closeSonatypeStagingRepository` is the close operation fails (due to some formal validation issues). He fixes the problem locally and call `publishToSonatype` again. The existing repo is found and the files are added to it (it wasn't closed due to the error). It's rather not the expected situation (and as I understood the latest code, it would behave that way).

I wonder if switching `initialize` to `CREATE_NEW` would "fix" the problem, it there are some corner cases? However, in that case (`CREATE_NEW`), I'm not sure if it still makes sense to merge the new logic into the `InitializeNexusStagingRepository` .

vlsi:
I am afraid this is getting nowhere. I would rather close the pr and fork gnpp

szpak:
> I am afraid we go in circles, and there's no way it could be merged. I would rather close the pr and fork gnpp

Sorry @vlsi, but this time I really don't get you.

Please take a look at the open comment yesterday:
- broken e2e tests - I "fixed" them partially and asked you about the idea how to fix the other part (the root project)
- a need for documentation

Otherwise, **I liked the version and wanted to merge it** (with working e2e tests and docs). Are those two comments above ridiculous?

---

And today, suddenly, you completely (really) reworked the whole concept. I put the two main comments:

- a common case which breaks the backward compatibility and - in my opinion - introduces a behavior that is not expected by the majority of people and ask for your comment:

> Talking about the feature itself, I might imagine a situation when, someone calls publishToSonatype and closeSonatypeStagingRepository is the close operation fails (due to some formal validation issues). He fixes the problem locally and call publishToSonatype again. The existing repo is found and the files are added to it (it wasn't closed due to the error). It's rather not the expected situation (and as I understood the latest code, it would behave that way).

- one too complex part of code - and here, sorry to say, but for me 5 levels of nested statements with complex logic inside was hard to read (maybe I'm just a Kotlin noob) and I proposed extracting it to smaller method:
image

I believe that for also for some other people reading that code it could be "not so straightforward" in that form.

If that "is getting nowhere", well, I don't know what to response.

vlsi:
What bothers me the most is:
* There's virtually no feedback on the high level
  * https://github.com/gradle-nexus/publish-plugin/pull/145#discussion_r1103829334
  * https://github.com/gradle-nexus/publish-plugin/issues/182 -- currently there's no way to execute tests, and there's no clarification why a part of the test code is moved to a different repository.
* There are a lot of comments on the style side like "extracting variables", "too long lines", "method names"

----

My expectations are:
* It is fine to comment on the code style in case the overall concept is ok
* It is not fine to ignore the high level and comment on variable names
* It is not fine to have "this breaks backward compatibility, and this variable has a bad name" at the same time

What I see is pure unpredictability. I have absolutely no idea what could be merged and what not.
I do not want to spend my time trying to figure out what pleases the maintainers of the project.

Of course, you are free to do whatever you want with the project, however, based on the reviews above I lost interest in contributing further.

szpak:
> https://github.com/gradle-nexus/publish-plugin/pull/145#discussion_r1103829334

You closed that one and in the other thread I [proposed](https://github.com/gradle-nexus/publish-plugin/pull/145#discussion_r1112304665) to generate a unique version in Gradle and call it twice in the e2e test (at least for single-module project it should work right now).

> https://github.com/gradle-nexus/publish-plugin/issues/182 -- currently there's no way to execute tests

That's not true. They are separate projects. You can clone it and release with a regular Gradle command from your console. Just provide your own staging profile and credentials. There were separated to have that freedom.
Similar approach (with own credentials) should be possible also to call them locally from e2e tests. Submodules can make it slightly less usable and we could consider their incorporation into the main repo, but it is not a blocker.

Having them (just the one with the single module, unless you already fixed the issue with the multi-module one) green locally (or feeling it should be), I can trigger them manually from CI, after the recent changes - #183. Neverhtless, if you feel overwhelmed by the submodules, I can do that myself in this branch. No problem. Just let me know.

szpak:
> Of course, you are free to do whatever you want with the project, however, based on the reviews above I lost interest in contributing further.

I see. In that case, I will probably take the previous version of this PR, which - as I wrote - is fine for me to merge (the one before you unexpectedly proposed revolutionary changes, which I had some doubts about and suggested an important case which [could be broken](https://github.com/gradle-nexus/publish-plugin/pull/145#issuecomment-1439148247), but you didn't even try to comment it), add one e2e test, write some documentation to allow people to find that feature and put it into the next release. Thanks for all your contribution (in code and comments) so far (in that project and many others!).

szpak:
Superseded by #201.

Issue #152, opened 2022-11-03
ConfigureUtil deprecation warning


ben-manes:
fyi I noticed this in a build scan,

> The org.gradle.util.ConfigureUtil type has been deprecated. 
This is scheduled to be removed in Gradle 9.0.
[Documentation](https://docs.gradle.org/7.6-rc-1/userguide/upgrading_version_7.html#org_gradle_util_reports_deprecations)

https://github.com/gradle-nexus/publish-plugin/blob/0aea95597e2eeaf9275a286254894bad29b3d1e1/src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt#L32

erichaagdev:
I've also started seeing this after upgrading to Gradle 7.6.

Build Scan: https://scans.gradle.com/s/ongbkxs4wvvba/deprecations?expanded=WyIzIl0&expanded-stacktrace=WyI1Il0&focused-stack-frame=0-0-2

remal:
Looks like `Project.configure()` method should be used instead of `ConfigureUtil.configureUsing()`

dili91:
I've also noticed the same since I bumped the wrapper to 7.6 inside one of my company's project 👀 

```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
```

If useful, [here](https://github.com/TrueLayer/truelayer-java/blob/main/build.gradle#L199) the `build.gradle` file

nedtwigg:
[8.0-rc-1 is out now](https://docs.gradle.org/8.0-rc-1/release-notes.html). This issue will become a blocker pretty soon.

TWiStErRob:
There's another callsite in the same file (curtesy of Mockito using this plugin):
```
The org.gradle.util.ConfigureUtil type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/7.6/userguide/upgrading_version_7.html#org_gradle_util_reports_deprecations
        at org.gradle.util.ConfigureUtil.logDeprecation(ConfigureUtil.java:216)
        at org.gradle.util.ConfigureUtil.configureSelf(ConfigureUtil.java:192)
        at io.github.gradlenexus.publishplugin.DefaultNexusRepositoryContainer.configure(DefaultNexusRepositoryContainer.kt:42)
[...]

PR #160, opened 2023-02-13
Fix deprecation warnings

nedtwigg:
This PR fixes the deprecation warnings noted in

- #152 

It does so by simply

- copy-pasting `org.gradle.util.ConfigureUtil` into the source of this project
- removing the `logDeprecation` calls
- removing all unused methods

This `ConfigureUtil` class adds about ~100 lines to the project. Unfortunately, even after merging this PR, there are still internal packages which Gradle may change in a future release. For this reason, I would recommend following Gradle's recommendation and removing the `Closure`-based API's in favor of only the `Action` API. I will submit a PR on top of this PR which shows that option.


nedtwigg:
`./gradlew compatTest` is now passing for me locally.

I maintain a few Gradle plugins myself, and I try quite hard to not bump the minimum requirements on any of them. However, that is trumped by "users should always be able to use the latest Gradle without warnings". If I need to bump the minimum requirements in order to make that possible, I do it. IMO this project ought to bump its minimum reqs, but with this PR you can delay that a bit longer if you'd like :)

src/main/java/io/github/gradlenexus/publishplugin/ClosureBackedAction.java
```diff
@@ -0,0 +1,108 @@
...
+import groovy.lang.Closure;
+import org.codehaus.groovy.runtime.metaclass.MissingMethodExecutionFailed;
+import org.gradle.api.Action;
+import org.gradle.api.InvalidActionClosureException;
+import org.gradle.util.Configurable;
+import java.util.Objects;
+
+/**
+ * Exact copy of [org.gradle.util.internal.ClosureBackedAction] from Gradle 7.6, except.
+
```
szpak, line 28: Extra line, maybe it breaks the javadoc generation on CI...

szpak:
> If I need to bump the minimum requirements in order to make that possible, I do it. IMO this project ought to bump its minimum reqs, but with this PR you can delay that a bit longer if you'd like :)

I still have some neglected plugins using the old `maven` module, but sometimes Gradle doesn't give you a clear way to support (too many) different versions :-/. If supporting Gradle 6 was problematic, we would switch to 7+. However, then I would like to publish a new version with some new features and bugfixes and the next one bumping the required Gradle versions. So, if keeping Gradle 6+ compatibility (for some time) is achievable, I would prefer to have it (especially that Gradle 8 has just been released and migration to it can take some time).

I propose to wait (a while) for any reply in https://github.com/gradle/gradle/issues/23874 before merging this PR.

nedtwigg:
All CI was failing with

```
> Task :javadoc FAILED
1 error
javadoc: error - No public or protected classes found to document.
```

I worked around this by making a single class public for no good reason. It can be removed completely when converting to `Action` and dropping `Closure`.

src/main/java/io/github/gradlenexus/publishplugin/ClosureBackedAction.java
nedtwigg: Just pushed a fix.

PR #161, opened 2023-02-13
Remove `Closure` methods and keep only `Action`

nedtwigg:
This PR fixes the deprecation warnings in

- #152

Instead of continuing to rely on Gradle internal APIs like the solution proposed in

- #160

This PR just removes the `Closure` methods entirely.


szpak:
This approach breaks compatibility with Groovy configuration. Let's cover it in #161 and #162.

PR #164, opened 2023-02-13
Refresh Stutter Locks

TWiStErRob:
```
git checkout master
git checkout -b stutter_locks
gradlew stutterWriteLocks
git add **/*.lock
git commit -m "Refresh Stutter Locks"
```


src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
```diff
@@ -949,10 +949,19 @@ class NexusPublishPluginTests {
         gradleRunner(*arguments).buildAndFail()
 
     private fun gradleRunner(vararg arguments: String): GradleRunner {
+        val warnings = when {
+            // There's a deprecation in Gradle 7.6 that doesn't allow for passing builds.
+            // See https://github.com/gradle-nexus/publish-plugin/issues/152 for more details.
+            // This conditional branch of when should be removed once that issue is fixed.
+            gradleVersion >= GradleVersion.version("7.6") -> "all"
```
TWiStErRob, line 956:
```suggestion
            gradleVersion >= GradleVersion.version("7.6") -> "fail"
```
would fail, because of the warning in https://github.com/gradle-nexus/publish-plugin/issues/152

Note: should there be a scheduled CI job, that runs every day/week which runs the commands in OP and opens a PR. This would've helped discover https://github.com/gradle-nexus/publish-plugin/issues/152 by automation.

TWiStErRob:
@szpak can we merge this before the fix for #152? So that testing for newer versions is covered. I'll remove the `fail`-`all` change in the PR for #152.

szpak:
> @szpak can we merge this before the fix for #152? So that testing for newer versions is covered. I'll remove the `fail`-`all` change in the PR for #152.

Sure, I was thinking about that, but... the warnings. However, as you suggested, they could be removed in the #152 related PR.

src/compatTest/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPluginTests.kt
TWiStErRob: Reverted in https://github.com/gradle-nexus/publish-plugin/pull/172/commits/3a607aa7496ea726e6edc3509d1fd396dc14ad98

TWiStErRob:
Interesting... I force pushed after you added to merge queue... (it was a clean rebase, but still, should've failed in some way; I wonder which version got merged)

szpak:
> Interesting... I force pushed after you added to merge queue... (it was a clean rebase, but still, should've failed in some way; I wonder which version got merged)

The text after I clicked the "merge when ready" stated that "any new commit will not be taken into account". However, it's still in beta ;).

Were there any new commits after that force-push 5 minutes before merge?

TWiStErRob:
No changes it was equivalent to the merge-queue because I just rebased.

PR #154, opened 2022-12-03
Update to Gradle 7.6

larsgrefer:

PR #168, opened 2023-02-16
Update from Gradle 6.9 to Gradle 8.0

TWiStErRob:
Closes #154

 * Including some necessary plugin updates.
 * Keeping Kotlin 1.3 as target compatibility.
 * Ensured that the generated artifacts are mostly the same.
   * Bytecode for most classes matches, where it doesn't I'm not sure what's different. There's a tiny diff in @kotlin.Metadata, probably a minor Kotlin 1.3.x diff.
   * MANIFEST.MF is the same
   * -sources.jar is the same
   * -javadoc.jar is still empty (😲)
   * Gradle plugin marker is correct
   * .pom files are the same (bar some expected trivial changes)
   * .module files are the same, bar some expected changes [(full diff)](https://user-images.githubusercontent.com/2906988/219407001-07537376-478b-40fa-87af-230d023bc93f.png):
     * `apiElements` and `runtimeElements` were removed


build.gradle.kts
```diff
@@ -5,12 +5,11 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     `kotlin-dsl`
```
vlsi, line 7: Technically speaking, you probably want replacing `kotlin-dsl` with `id("...")` so Gradle 8 improved `plugins interpreter` could kick-in here

>Gradle 8.0 introduces an interpreter for the declarative plugins {} blocks in .gradle.kts scripts that make the overall build time around 20% faster. Calling the Kotlin compiler for declarative plugins {} blocks is avoided

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-all.zip
```
vlsi, line 3: Should it be `8.0-bin` to save traffic on CI builds?

```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
```

src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
```diff
@@ -20,16 +20,14 @@ import groovy.lang.Closure
...
 import java.net.URI
 import javax.inject.Inject
 
-@Suppress("UnstableApiUsage")
 internal open class DefaultNexusRepositoryContainer @Inject constructor(
     delegate: NamedDomainObjectContainer<NexusRepository>
 ) : NexusRepositoryContainer, NamedDomainObjectContainer<NexusRepository> by delegate {
 
-    override fun sonatype(): NexusRepository = sonatype {}
+    override fun sonatype(): NexusRepository = sonatype(Action {})
```
vlsi, line 30: Is this change needed?

gradle/wrapper/gradle-wrapper.properties
TWiStErRob: It saves on CI builds, but seriously degrades local developer experience (code navigation). Especially that this is a Gradle plugin :)

build.gradle.kts
TWiStErRob: Nice, except the version must be specified in that case.

```
Plugin [id: 'org.gradle.kotlin.kotlin-dsl'] was not found in any of the following sources:
Plugin [id: 'kotlin.kotlin-dsl'] was not found in any of the following sources:
Plugin [id: 'kotlin-dsl'] was not found in any of the following sources:

- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Plugin Repositories (plugin dependency must include a version number for this source)
```

We could hard-code 4.0.6, which is generated as 
```kotlin
val PluginDependenciesSpec.`kotlin-dsl`: PluginDependencySpec
    get() = id("org.gradle.kotlin.kotlin-dsl") version "4.0.6"
```
from https://github.com/gradle/gradle/blob/v8.0.0/build-logic/kotlin-dsl/src/main/kotlin/gradlebuild/kotlindsl/generator/tasks/GenerateKotlinDependencyExtensions.kt

See also
https://github.com/gradle/gradle/blob/v8.0.0/subprojects/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt#L56-L65

TWiStErRob: They did it somehow here:
https://github.com/gradle/gradle/blob/v8.0.0/build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.build-logic.kotlin-dsl-gradle-plugin.gradle.kts

TWiStErRob: According to https://docs.gradle.org/8.0/userguide/plugins.html#sec:constrained_syntax, `kotlin-dsl` should be accepted special case...

vlsi: indeed, it should be a special case

build.gradle.kts
```diff
@@ -68,21 +65,18 @@ idea {
...
 configurations {
-    compileOnly {
-        extendsFrom(shadowed)
-    }
     testImplementation {
-        extendsFrom(shadowed)
         exclude(group = "junit", module = "junit")
     }
+    // Workaround https://github.com/gradle/gradle/issues/23928
+    shadow.configure { afterEvaluate { this@configure.dependencies.remove(project.dependencies.gradleApi()) } }
```
vlsi, line 73: I just verified, and this "workaround" does not really help. Is it required?
What helps is that the current `build.gradle.kts` configures manifest for `shadowJar` tasks, and it somehow removes `Class-Path` attribute.

gradle/wrapper/gradle-wrapper.properties
vlsi: >It saves on CI builds, but seriously degrades local developer experience (code navigation)

Can you elaborate which code navigation it breaks?
IDEA resolves and downloads Gradle sources just fine even if you mention `-bin` distribution.

src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
vlsi: It would save a WTF moment if you added that into the code comment 😉

build.gradle.kts
TWiStErRob: Huh? When I comment it out, the line appears, otherwise it's not there.

TWiStErRob: > it somehow removes Class-Path attribute.

I think that might be because it throws away the `shadow` configuration in its entirety, fully reconfiguring the shadow plugin from scratch.

TWiStErRob: Let's see https://github.com/gradle-nexus/publish-plugin/pull/168/commits/c14acc5dcaa8a9ff360d60aaf89f4965d109a189

src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
TWiStErRob: I think it's ok if you're not looking at the diff. And if you try to "fix" it, you'll get a compile error. What would be the comment?

gradle/wrapper/gradle-wrapper.properties
TWiStErRob: Yeah, I've heard that a lot, but then I wasn't able to find/open classes from Gradle in a project which used -bin. I'll try to find an example.

src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
vlsi: For instance:

```
// sonatype {} here does not work with Kotlin 1.3 for unknown reason
```

build.gradle.kts
```diff
@@ -141,21 +135,34 @@ sourceSets {
...
+            // but sadly the @kotlin.Metadata created on the classes would be incompatible with Kotlin 1.3 consumers.
+            kotlinOptions.languageVersion = "1.3"
+            doFirst {
+                if (kotlinOptions.apiVersion == "1.3") {
+                    // Suppress "Language version 1.3 is deprecated and its support will be removed in a future version of Kotlin".
+                    kotlinOptions.freeCompilerArgs += "-Xsuppress-version-warnings"
+                } else {
+                    TODO("Remove -Xsuppress-version-warnings suppression, or change the condition to ${kotlinOptions.languageVersion}")
+                }
+            }
```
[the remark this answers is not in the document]
TWiStErRob: Which part do you mean? 😅

 * I got `afterEvaluate { kotlinOptions }` simply by reading the code after being annoyed that without `afterEvaluate` it doesn't do anything.
 * The `doFirst` is something I started doing recently (a year or so), so that any hacks will auto-remove themselves when they're not necessary any more.
 * `apiVersion`/`languageVersion` is not a hack, the project has old compatible consumers, so need to back-compile.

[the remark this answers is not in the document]
TWiStErRob: I recommend looking at individual commits, for example this built up in 3:
https://github.com/gradle-nexus/publish-plugin/pull/168/commits/c63aa3bb651ab438b2d7e9b08d70311d693336fc
https://github.com/gradle-nexus/publish-plugin/pull/168/commits/6f8bdfaa43b2a6808032509c8fcdb8cfd14c7fe2
https://github.com/gradle-nexus/publish-plugin/pull/168/commits/dd958d126e3e30ae0049a891f7a21aeb4915993f

build.gradle.kts
vlsi: I downloaded `ubuntu-latest build folder.zip` (sha `ecc6e7f77e93cdc6e300595c1cfa33460b857079`)
`libs/publish-plugin-1.1.1-SNAPSHOT.jar` (sha `8ab95464dd7954b278ecf63338cb8ab55c3c4e47`)

Manifest contents is

```
Manifest-Version: 1.0
Implementation-Version: 1.1.1-SNAPSHOT
```

TWiStErRob: I was a bit too bold with uploading the build/ folder, it contains the testkit folders 😅

https://github.com/gradle-nexus/publish-plugin/actions/runs/4204607452
(don't be scared about 952M, the compressed version is only 300M)

Anyway, `windows-latest build folder.zip/libs/publish-plugin-1.1.1-SNAPSHOT.jar` contains no .jars listed in the manifest, so the workaround works.

src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt
TWiStErRob: https://github.com/gradle-nexus/publish-plugin/pull/168/commits/20f22c49c56f994ba839eede2befd54bc75d84cd

TWiStErRob: Also I think it's exactly because of this line:
https://github.com/gradle/gradle/blob/v8.0.0/subprojects/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslCompilerPlugins.kt

which was different in

https://github.com/gradle/gradle/blob/v6.9.0/subprojects/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslCompilerPlugins.kt#L61

TWiStErRob: I tried enabling new inference with: `kotlinOptions.freeCompilerArgs += "-XXLanguage:+NewInference"`, but I still get an error and a huge warning:
```
> Task :compileKotlin FAILED
w: ATTENTION!
This build uses unsafe internal compiler arguments:

-XXLanguage:+NewInference

This mode is not recommended for production use,
as no stability/compatibility guarantees are given on
compiler or generated code. Use it at your own risk!

e: file:///P:/projects/contrib/github-nexus-publish-plugin/src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt:30:57 Type mismatch: inferred type is () -> Unit but Action<in NexusRepository> was expected

FAILURE: Build failed with an exception.
```
So lets stick with these two tiny syntax changes with comments.

build.gradle.kts
vlsi: I mean my local build (e.g. after removing `.gradle` and `build`) does not contain `class-path` attribute with the following diff:

```diff
diff --git a/build.gradle.kts b/build.gradle.kts
index 83ab07e..b96e667 100644
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -70,7 +70,7 @@ configurations {
         exclude(group = "junit", module = "junit")
     }
     // Workaround https://github.com/gradle/gradle/issues/23928
-    shadow.configure { afterEvaluate { this@configure.dependencies.remove(project.dependencies.gradleApi()) } }
+//    shadow.configure { afterEvaluate { this@configure.dependencies.remove(project.dependencies.gradleApi()) } }
 }

 dependencies {
diff --git a/gradle/wrapper/gradle-wrapper.properties b/gradle/wrapper/gradle-wrapper.properties
index 2848974..42defcc 100644
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
 networkTimeout=10000
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
```

gradle/wrapper/gradle-wrapper.properties
TWiStErRob: Here's what I did:
 * `8.0-all` (current branch state) -> navigation works
 * change `8.0-all` to `8.0-bin`
 * sync -> navigation still works,  
   but source code is loaded from `-all` folder in `wrapper/dists`
 * sync, restart, sync -> navigation still works,  
   but source code is loaded from `-all` folder in `wrapper/dists`
 * change `8.0-bin` to `8.0-rc-4-bin` to simulate a "fresh start"
 * sync -> I saw `gradle-8.0-rc-4-src.zip` (40M) downloaded -> navigation works
   this is when I started really testing

Here are some of the use cases that I tested apart from simple Ctrl+Click.

1. See all the files in a specific .jar file that contains the class:
   Works the better in `-all`.
   * navigate to class (e.g. `NamedDomainObjectContainer`)
   * press "Select Opened File" ("target" button) on Project view.
   * go to .jar's root and drill down to META-INF and other files/folders
   * when I go up to the .jar's root I see this:
     image
   * At least in `-all` it shows the folder names:
     image
2. Navigate to a specific plugin by ID: Navigate > File ... > `org.gradle.java.properties`
   Works in both, except with `-bin` all the files are duplicated:
     image
3. Focusing on any particular class doesn't show which .jar it belongs to  
   broken in both, this only works for normal dependencies
   * open some source code of any Gradle class
   * press "Select Opened File" ("target" button) on Project view.
   * Observe location
     * bin: image
     * all: image
4. Opening a .class file via navigate to file, e.g. `NamedDomainObjectContainer.class`
   (this happens sometimes automatically when debugging, probably more with Kotlin):
   broken in both, unless you explicitly attach the source code to the .class files.
    ```
    //
    // Source code recreated from a .class file by IntelliJ IDEA
    // (powered by FernFlower decompiler)
    //
    ```
    or even worse
    ```
    // IntelliJ API Decompiler stub source generated from a class file
    // Implementation of methods is not available
    
    package org.gradle.kotlin.dsl
    ... { /* compiled code */ }
    ```

So I'm not entirely sure what didn't work still. `-all` has a little bit better experience in advanced use cases, but generally looks the same, so [I'm lowering the load](https://github.com/gradle-nexus/publish-plugin/pull/168/commits/45afbcd498be398026dcb0a65e7fcae70d46c513); and I'll vigilantly watch out for WTFs related to Gradle source navigation.

vlsi: The duplicated `java` folder names look bizzare. It is probably worth creating a YouTrack issue for that. I've no idea how to search the one, so I would probably just file a new one.

vlsi (approved):
This looks good to me.
I especially like the way you structure commits and add relevant comments pointing to issues, and the reasons.

build.gradle.kts
TWiStErRob: Attachments on https://github.com/gradle-nexus/publish-plugin/actions/runs/4206302177 show that the classpath is generated without the workaround.

TWiStErRob:
@szpak this is RTM as you see fit.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
```
szpak, line 3: Could you add `distributionSha256Sum`? E.g. with (for Fish shell):

```
GRADLE_VERSION=8.0 gw wrapper --gradle-version $GRADLE_VERSION --gradle-distribution-sha256-sum (wget -O- -q "https://services.gradle.org/distributions/gradle-$GRADLE_VERSION-bin.zip.sha256")
```

And "vote" for the built-in task in Gradle: https://github.com/gradle/gradle/issues/12785 ;-)

szpak (changes requested):

szpak:
Thanks @TWiStErRob. And thanks @vlsi for the review. Just one nitpicking and we can merge it.

gradle/wrapper/gradle-wrapper.properties
TWiStErRob: Doing. Is this a secondary check in addition to "validate wrapper"?

szpak: It's mostly for the local execution. To detect corrupted binaries (or some kind of MiTM attacks). The "validate wrapper" action is even better, as it (AFAIR) compares the actual checksum with the record published by Gradle.

vlsi: It is to validate the checksum on developer's machines.

Frankly speaking, I thought of asking the checksum, however, it was not there before, so I skipped

TWiStErRob: I'm on Windows; I downloaded it manually and copy-pasted from
https\://services.gradle.org/distributions/gradle-8.0-bin.zip.sha256

https://github.com/gradle-nexus/publish-plugin/pull/168/commits/2cecfc0b193f1fe6cd362db6cdfc4dff19448526

szpak: Alternatively, they are on: https://gradle.org/release-checksums/

szpak (approved):

TWiStErRob:
... and of course 1 hour later, Gradle 8.0.1 is out. 🙃

Issue #171, opened 2023-02-16
Deprecate support for Java 8 (in fact <11)


szpak:
Having 1.2.0 released (one day ;-) ) we could consider deprecating the (official) support for JDK <11. As the project uses Kotlin, the syntax is not a big deal, but Java 8 is very old and we could help convince people to migrate. In the first phase, we could just remove JDK 8 build from CI, but keep source/target compatibility with JDK 8, unless some our dependencies are only JDK 11+.

As the java toolchain functionality is available in Gradle, people could still compile their code with JDK 8, if really needed (even if we switch hard to JDK 11).

Feel free to comment (or "vote") here, if you have any strong arguments against that idea (or even if you would like to make a hard switch to JDK 11, not to mention 17 ;-) ).

vlsi:
If the project is built with target 8, that is fine to me.
It would be sad if the plugin required Java 11 for Gradle execution.

TWiStErRob:
> It would be sad if the plugin required Java 11 for Gradle execution.

Agree, until Gradle drops Java 8 support, all plugin authors (looking at you AGP 😉), should keep that minimum version.

> If the project is built with target 8, that is fine to me.

Keeping target 8 is fine, it means the produced JAR will be usable with Java 8 or anyone using target 8 in their builds. At this point it doesn't matter at all which JDK you use, you can even build on JDK 19 as long as your target is 8.

I would like to mention here that there's a common misconception of CI matrixes, really well explained here: https://jakewharton.com/build-on-latest-java-test-through-lowest-java/

vlsi:
I incline https://github.com/vlsi/github-actions-random-matrix is a better approach for CI matrix.

See https://twitter.com/vladimirsitnikv/status/1623265219383263232

What do you think of switching to a randomized matrix?

TWiStErRob:
The point of the link is that you don't even need a CI matrix, build the thing once on the latest version, and then run the tests on different JREs. I agree with randomizing parameters **after** the simplification has been made.

vlsi:
> you don't even need a CI matrix

I'm not sure how are you going to test Windows/Linux without a matrix 😀

TWiStErRob:
Fair, I'm referring to Java versions... 😅

https://github.com/gradle-nexus/publish-plugin/blob/54dff4e09ceaffb038348972ef6e9fecb1e53990/.github/workflows/java-versions.yml#L15-L20

ciscoo:
[...]

PR #174, opened 2023-02-18
Cron tests with latest Gradle version(s) on CI

### LINKED ISSUE #173: Nightly (weekly) compatibility check with the latest Gradle version
A generic idea, recently re-proposed also by @TWiStErRob to run basic compatibility checks (e.g. fail with deprecations) with the latest Gradle versions (stable + RC + milestones or also with snapshots?) on a regular basis (weekly?). In fact, currently we have the weekly "cron" bulids to run e2e-tests with Nexus, there are could also called a separate GH action (to keep our generic builds somehow reproducible) with the latest Gradle version.

The basic idea could be to append the latest Gradle version (determined using [this API](https://services.gradle.org/versions/)) to `.shutter/java17.lock` at the beginning the of build. There could be a dedicated plugin or maybe a dedicated GH action to achieve that.

szpak:
To faster detect regressions, deprecations and incompatibilities. Simple variant to get started.

Closes #173.

### DIFF
--- a/.github/workflows/gradle-latest-versions.yml
+++ b/.github/workflows/gradle-latest-versions.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   gradle-compatTest:
-    name: "Compatibility tests for latest Gradle versions"
+    name: "Compatibility tests with latest Gradle versions"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -23,7 +23,7 @@ jobs:
         run: |
           echo -n "" > .stutter/java11.lock
           for gv in current release-candidate nightly release-nightly; do
-            echo "Fetching and setting Gradle version for $gv"
+            echo "Fetching and setting Gradle version for '$gv'"
             wget -O- -q https://services.gradle.org/versions/$gv | jq -r '.version | select( . != null )' >> .stutter/java11.lock
           done
           echo "Configured Gradle versions for compatibility tests:"
--- a/.github/workflows/gradle-latest-versions.yml
+++ b/.github/workflows/gradle-latest-versions.yml
@@ -3,9 +3,6 @@ name: Gradle latest versions tests
 on:
   schedule:
     - cron: "0 7 * * MON"
-  push:
-    branches:
-      - "szpak/gradleLatestVersionsGHActions"
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/gradle-latest-versions.yml
+++ b/.github/workflows/gradle-latest-versions.yml
@@ -6,8 +6,8 @@ on:
     - cron: "0 7 * * MON"
   push:
     branches:
-      - "gradleLatestVersionsTestManualTrigger"
       - "szpak/gradleLatestVersionsGHActions"
+  workflow_dispatch:
 
 jobs:
   gradle-compatTest:
--- a/.github/workflows/gradle-latest-versions.yml
+++ b/.github/workflows/gradle-latest-versions.yml
@@ -1,4 +1,3 @@
-# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
 name: Gradle latest versions tests
 
 on:
--- a/.github/workflows/gradle-latest-versions.yml
+++ b/.github/workflows/gradle-latest-versions.yml
@@ -1,0 +1,32 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: Gradle latest versions tests
+
+on:
+  schedule:
+    - cron: "0 7 * * MON"
+  push:
+    branches:
+      - "gradleLatestVersionsTestManualTrigger"
+      - "szpak/gradleLatestVersionsGHActions"
+
+jobs:
+  gradle-compatTest:
+    name: "Compatibility tests for latest Gradle versions"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          java-version: 11
+          distribution: 'zulu'
+      - name: Fetch and set latest Gradle versions to test with
+        run: |
+          echo -n "" > .stutter/java11.lock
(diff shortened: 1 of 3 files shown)


.github/workflows/gradle-latest-versions.yml
```diff
@@ -0,0 +1,32 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: Gradle latest versions tests
+
+on:
+  schedule:
+    - cron: "0 7 * * MON"
+  push:
+    branches:
+      - "gradleLatestVersionsTestManualTrigger"
+      - "szpak/gradleLatestVersionsGHActions"
```
szpak, line 10: The first branch is to be able to retrigger it on demand. The second will be removed before merge.

Do we need any "other" mode of execution?

.github/workflows/gradle-latest-versions.yml
```diff
@@ -0,0 +1,45 @@
...
+
+          for gv in $GRADLE_VERSIONS_TYPES_TO_FETCH; do
+            echo "Fetching latest Gradle version for '$gv'"
+            curl --silent --show-error "https://services.gradle.org/versions/$gv" | jq -r '.version | select( . != null )' >> latest-gradle-versions.txt
+          done
+
+          echo -e "\nGradle versions configured for compatibility tests (in $GRADLE_VERSIONS_OUTPUT_FILE):"
+          cat $GRADLE_VERSIONS_OUTPUT_FILE
+
+          echo "GRADLE_VERSIONS_OUTPUT_FILE=$GRADLE_VERSIONS_OUTPUT_FILE" >> $GITHUB_ENV
```
szpak, line 36: Here it is mostly meaningless, but in the context of (potentially) externalize it to a separate shared action, I wonder, if setting the versions itself (with some "safe" separator) woudn't be a better option?

.github/workflows/gradle-latest-versions.yml
vlsi: I guess "manual" trigger should better be configured via `workflow_dispatch`.
Then you have a GitHub UI for triggering a run, and you can even have parameters (e.g. a Gradle version to test): https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

szpak: It makes sense. I've been doing something similar with the variables in GitLab.

.github/workflows/gradle-latest-versions.yml
```diff
@@ -0,0 +1,45 @@
...
+          distribution: 'zulu'
+
+      - name: Fetch latest available Gradle versions
+        run: |
+          GRADLE_VERSIONS_TYPES_TO_FETCH="current release-candidate nightly release-nightly"
+          GRADLE_VERSIONS_OUTPUT_FILE=latest-gradle-versions.txt
+
+          for gv in $GRADLE_VERSIONS_TYPES_TO_FETCH; do
+            echo "Fetching latest Gradle version for '$gv'"
+            curl --silent --show-error "https://services.gradle.org/versions/$gv" | jq -r '.version | select( . != null )' >> latest-gradle-versions.txt
```
TWiStErRob, line 30: Isn't `gradlew stutterRefreshLocks` giving you similar results?

szpak: Could you elaborate more? I've checked the https://github.com/ajoberstar/gradle-stutter repo, but I [searching for "refresh"](https://github.com/ajoberstar/gradle-stutter/search?q=refresh) did not return anything :-/.

vlsi: https://github.com/ajoberstar/gradle-stutter/blob/709895b0061354ef5999f32444652e9f10b7bf94/src/main/java/org/ajoberstar/gradle/stutter/StutterPlugin.java#L35-L38 ?

TWiStErRob: Yep that, sorry.

szpak: 1. We don't use the latest version and I'm not sure if it wasn't added in 0.7.x (which [dropped support for JDK 8](https://github.com/ajoberstar/gradle-stutter/releases/tag/0.7.0)).
2. I'm not sure about the nightly version available there (but maybe we don't need it?)
3.

.github/workflows/gradle-latest-versions.yml
szpak: I added `on.workflow_dispatch` which should be enough, but it's not :-).

image

In GitLab, I could select the branch to run workflow with, here, I don't see anything similar. On the main branch, there isn't "workflow_dispatch", maybe [because of that](https://stackoverflow.com/a/65389878) the GH Actions don't allow mi to run it manually... (but will, after it is merged into master). I don't have also GH cli configured to trigger it manually.

.github/workflows/gradle-latest-versions.yml
TWiStErRob: 1. It still has that task, that's how I did #164. My point here was that rather than manually finding the latest, just run that task and then run the builds. It'll pick up the latest Gradle versions anyway.
2. Yes nightlys won't be there, but that's not necessary anyway, because you would be exposed to potentially buggy Gradle builds resulting in flaky CI builds every Monday, which is pretty random. I wonder if you can tell Stutter to use RCs though.

szpak: I propose to merge it as is and keep it running (the next window is ~12h), until we switch to building with [JDK 11+](https://github.com/gradle-nexus/publish-plugin/issues/171), bump the Stutter and find a way to use also RC builds.

TWiStErRob: All yours, was just raising a potential issue.

.github/workflows/gradle-latest-versions.yml
szpak: Having it merged into master, there is a button:
image

Issue #228, opened 2023-06-02
Failing "IvyNexusPublishPluginTests > can be used with lazily applied Gradle Plugin Development Plugin" on CI with LockTimeoutException


szpak:
The (compatibility) test "IvyNexusPublishPluginTests > can be used with lazily applied Gradle Plugin Development Plugin" failed twice on CI ([1](https://github.com/gradle-nexus/publish-plugin/actions/runs/5152744338/jobs/9279172923#step:4:216), [2](https://github.com/gradle-nexus/publish-plugin/actions/runs/5152865393/jobs/9280072186#step:4:222)) in short period of time. Both times on Mac and both times in the Ivy version (once with 7.6.1, once with 8.0.2) . The error in the output is not very helpful:

```
IvyNexusPublishPluginTests > can be used with lazily applied Gradle Plugin Development Plugin FAILED
    org.gradle.testkit.runner.UnexpectedBuildFailure at DefaultGradleRunner.java:269
```

luckily, the [build scan](https://gradle.com/s/kgpvb2h4jwfbi) is more verbose:
```
> Task :initializeSonatypeStagingRepository	
Created staging repository 'orgexample-42' at http://localhost:49978/repositories/orgexample-42/content/	
	
> Task :gradle-plugin:generateDescriptorFileForFooPluginMarkerIvyPublication	
> Task :gradle-plugin:publishFooPluginMarkerIvyPublicationToSonatypeRepository	
> Task :gradle-plugin:generateDescriptorFileForPluginIvyPublication	
> Task :gradle-plugin:compileJava	
> Task :gradle-plugin:compileJava FAILED	
	
FAILURE: Build failed with an exception.	
	
* What went wrong:	
Execution failed for task ':gradle-plugin:compileJava'.	
> Could not resolve all files for configuration ':gradle-plugin:compileClasspath'.	
   > Timeout waiting to lock Generated Gradle JARs cache (/Users/runner/work/publish-plugin/publish-plugin/build/tmp/compatTest7.6.1/work/.gradle-test-kit/caches/7.6.1/generated-gradle-jars). It is currently in use by another Gradle instance.	
     Owner PID: 26887	
     Our PID: 26886	
     Owner Operation: 	
     Our operation: 	
     Lock file: /Users/runner/work/publish-plugin/publish-plugin/build/tmp/compatTest7.6.1/work/.gradle-test-kit/caches/7.6.1/generated-gradle-jars/generated-gradle-jars.lock	
	
[...]

PR #231, opened 2023-06-08
Add manual invalidation of IvyRepository urls

maczikasz:
In gradle 8.2 the url became the part of the artifactResourcePatterns are stored with the URL, and the URL is taken from the repository description RepositoryDescriptions are created when the IvyArtifactRepository is created. The nexus plugin uses a deferred creation of staging repositories and relies on a lazy provider to provide the staging repository url to the artifact repository This no longer works as the url is no longer read at any later point.

This PR will call the `invalidateDescriptor()` method on the IvyArtifactRepository via reflection if possible.


src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    fun tryGet(name: String) = mapping[name]
+
+    override fun toString() = mapping.toString()
+
+    companion object {
+        private val log: Logger = LoggerFactory.getLogger(BasicActionRetrier::class.java)
+
+        private fun ArtifactRepository.invalidate() {
```
szpak, line 48: Could add a comment here about being "a workaround" and preferably with a link to a Gradle issue on GitHub to consider a better possibility to handle that? Or assuming `InvalidatingStagingRepositoryDescriptorRegistry` extends `StagingRepositoryDescriptorRegistry`, the comment could be made on top of the new class.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun invalidateLater(name: String, artifactRepository: ArtifactRepository) {
+        invalidateMapping[name] = artifactRepository
+    }
+
+    fun tryGet(name: String) = mapping[name]
+
+    override fun toString() = mapping.toString()
+
+    companion object {
+        private val log: Logger = LoggerFactory.getLogger(BasicActionRetrier::class.java)
```
szpak, line 46:
```suggestion
        private val log: Logger = LoggerFactory.getLogger(InvalidatingStagingRepositoryDescriptorRegistry::class.java)
```
?

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
```diff
@@ -0,0 +1,60 @@
...
+
+package io.github.gradlenexus.publishplugin.internal
+
+import org.gradle.api.artifacts.repositories.ArtifactRepository
+import org.gradle.api.internal.artifacts.repositories.AbstractResolutionAwareArtifactRepository
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import java.util.concurrent.ConcurrentHashMap
+
+class InvalidatingStagingRepositoryDescriptorRegistry {
```
szpak, line 25: I wonder, if `InvalidatingStagingRepositoryDescriptorRegistry` couldn't extend `StagingRepositoryDescriptorRegistry` (unless there are any technical issues with that)?

Assuming this is a temporary workaround, it would be easier to revert that in the future (currently, GitHub reports this class as a new one, and `StagingRepositoryDescriptorRegistry` as deleted, what can lead to harder access to the history of changes).

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -281,12 +282,14 @@ class NexusPublishPlugin : Plugin<Project> {
...
-        provideFallback: Boolean
+        registry: Provider<InvalidatingStagingRepositoryDescriptorRegistry>,
+        provideFallback: Boolean,
+        artifactRepo: ArtifactRepository
     ) = if (extension.useStaging.get()) {
         val descriptorRegistry = registry.get()
         if (provideFallback) {
-            descriptorRegistry.tryGet(nexusRepo.name)?.stagingRepositoryUrl ?: nexusRepo.nexusUrl.get()
+            descriptorRegistry.invalidateLater(nexusRepo.name, artifactRepo)
+            descriptorRegistry.tryGet(nexusRepo.name)?.stagingRepositoryUrl ?: URI(nexusRepo.nexusUrl.get().toString() + "/breakme/")
```
szpak, line 292: > "/breakme/"

Has it left from your tests?

It would mean that we don't have a test to cover not set `stagingRepositoryUrl` :-/.

src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt
```diff
@@ -52,6 +52,7 @@ open class InitializeNexusStagingRepository @Inject constructor(
         val descriptor = client.createStagingRepository(stagingProfileId, repositoryDescription.get())
         val consumerUrl = HttpUrl.get(serverUrl)!!.newBuilder().addEncodedPathSegments("repositories/${descriptor.stagingRepositoryId}/content/").build()
         logger.lifecycle("Created staging repository '{}' at {}", descriptor.stagingRepositoryId, consumerUrl)
+
         registry.get()[repository.name] = descriptor
```
szpak, line 56: An accidental empty line to be reverted?

szpak (changes requested):
That's good that you were able to find a workaround. I put some small remarks inline.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
maczikasz: Copy paste error sorry

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
maczikasz: Yap agree, makes sense to extend the StagingRepositoryDescriptorRegistry and add the comment

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import java.util.concurrent.ConcurrentHashMap
+
+/*
+ * This class is a temporary workaround to invalidate the repository description on a IvyArtifactRepository after the creation of the staging sonatype repo
+ */
+class InvalidatingStagingRepositoryDescriptorRegistry : StagingRepositoryDescriptorRegistry() {
+
+    private val mapping = ConcurrentHashMap<String, StagingRepositoryDescriptor>()
```
szpak, line 30: Do we need `mapping` here? `toString()` which seems to be the only to use it, could be reused from the super class.

szpak (changes requested):

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
maczikasz: no, not at all

szpak:
> spotless..

You are not the only one annoyed by the spotless failures on CI :-D

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
maczikasz: This was part of my debugging to figure out wether this is related, but good point about manually setting the staging repo url to null or not setting it at all, in that case we would try to use the nexusUrl as fallback

szpak:
It seems that the macos-latest build will fail due to #228 . This PR can be merged anyway, but do you know why the timeout happens with the Ivy compatibility tests?

maczikasz:
> It seems that the macos-latest build will fail due to #228 . This PR can be merged anyway, but do you know why the timeout happens with the Ivy compatibility tests?

no clue, did not see this locally I can ask around

szpak (approved):

szpak:
With your change, the test are [green again](https://github.com/gradle-nexus/publish-plugin/actions/runs/5211983504/jobs/9404977715#step:6:65) with the latest Gradle snapshot versions, thanks!

PR #234, opened 2023-06-29
Fix sample gradle files

espertus:
Prior versions gave me this error:

> Cannot create a Publication named 'pom' because this container does not support creating elements by name alone. Please specify which subtype of Publication to create. Known subtypes are: MavenPublication

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,28 +227,29 @@ publishing {
     publications {
         mavenJava(MavenPublication) {
             from(components.java)
-        }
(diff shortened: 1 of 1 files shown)


szpak:
Thanks. It might work in the past or I could make a mistake creating the examples :).

codeconsole:
Please merge this. I wasted a few hours trying to figure out what I was doing wrong.

PR #241, opened 2023-07-05
Partially refactor to use more idiomatic Gradle

3flex:
The main changes are:
1. Use managed properties, where possible
2. Avoid passing instances of `Project` around
3. Use the plugin to set default values in tasks and the extension

Parts 1 & 3 are not fully completed but I didn't want to press on if this wasn't desired. This should be safe to merge as is though as compat tests for Gradle 6 on Java 11 pass locally.

### DIFF
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
@@ -19,31 +19,33 @@ package io.github.gradlenexus.publishplugin
 import org.gradle.api.Action
 import org.gradle.api.NamedDomainObjectFactory
 import org.gradle.api.Project
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Nested
 import org.gradle.kotlin.dsl.container
 import org.gradle.kotlin.dsl.newInstance
-import org.gradle.kotlin.dsl.property
 import java.time.Duration
 
-open class NexusPublishExtension(project: Project) {
+abstract class NexusPublishExtension(project: Project) {
 
     companion object {
         internal const val NAME = "nexusPublishing"
     }
 
-    val useStaging = project.objects.property<Boolean>()
+    abstract val useStaging: Property<Boolean>
 
-    val packageGroup = project.objects.property<String>()
+    abstract val packageGroup: Property<String>
 
-    val repositoryDescription = project.objects.property<String>()
+    abstract val repositoryDescription: Property<String>
 
     // staging repository initialization can take a few minutes on Sonatype Nexus
-    val clientTimeout = project.objects.property<Duration>()
+    abstract val clientTimeout: Property<Duration>
 
-    val connectTimeout = project.objects.property<Duration>()
+    abstract val connectTimeout: Property<Duration>
 
-    val transitionCheckOptions = project.objects.property<TransitionCheckOptions>()
+    @get:Nested
+    abstract val transitionCheckOptions: TransitionCheckOptions
 
-    fun transitionCheckOptions(action: Action<in TransitionCheckOptions>) = action.execute(transitionCheckOptions.get())
+    fun transitionCheckOptions(action: Action<in TransitionCheckOptions>) = action.execute(transitionCheckOptions)
 
     val repositories: NexusRepositoryContainer = project.objects.newInstance(
         DefaultNexusRepositoryContainer::class,
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
@@ -30,24 +30,18 @@ open class NexusPublishExtension(project: Project) {
         internal const val NAME = "nexusPublishing"
     }
 
-    val useStaging = project.objects.property<Boolean>().apply {
-        set(project.provider { !project.version.toString().endsWith("-SNAPSHOT") })
-    }
+    val useStaging = project.objects.property<Boolean>()
 
-    val packageGroup = project.objects.property<String>().apply {
-        set(project.provider { project.group.toString() })
-    }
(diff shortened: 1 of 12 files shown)


3flex:
Looks like failure is due to #228

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -68,7 +82,20 @@ class NexusPublishPlugin : Plugin<Project> {
     }
 
     private fun configureNexusTasks(rootProject: Project, extension: NexusPublishExtension, registry: Provider<InvalidatingStagingRepositoryDescriptorRegistry>) {
+        rootProject.tasks.withType(AbstractNexusStagingRepositoryTask::class.java) {
```
TWiStErRob, line 85: Do we need configureEach here?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
```diff
@@ -18,48 +18,45 @@ package io.github.gradlenexus.publishplugin
...
+    abstract val transitionCheckOptions: TransitionCheckOptions
 
-    fun transitionCheckOptions(action: Action<in TransitionCheckOptions>) = action.execute(transitionCheckOptions.get())
+    fun transitionCheckOptions(action: Action<in TransitionCheckOptions>) = action.execute(transitionCheckOptions)
 
-    val repositories: NexusRepositoryContainer = project.objects.newInstance(
+    val repositories: NexusRepositoryContainer = objects.newInstance(
         DefaultNexusRepositoryContainer::class,
-        // `project.container(NexusRepository::class) { name -> ... }`,
+        // `objects.domainObjectContainer(NexusRepository::class) { name -> ... }`,
```
TWiStErRob, line 53: Ooh, nice one! I was looking for how to replace this!

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
```diff
@@ -18,48 +18,45 @@ package io.github.gradlenexus.publishplugin
...
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Nested
+import org.gradle.kotlin.dsl.domainObjectContainer
 import org.gradle.kotlin.dsl.newInstance
-import org.gradle.kotlin.dsl.property
 import java.time.Duration
+import javax.inject.Inject
 
-open class NexusPublishExtension(project: Project) {
+abstract class NexusPublishExtension @Inject constructor(objects: ObjectFactory) {
```
TWiStErRob, line 29: Hey @3flex! I did a similar migration as a try to fix an issue, check out this branch:
https://github.com/TWiStErRob/nexus-publish-gradle-plugin/commits/configuration-cache-induced-idiomatic-gradle-improvements

You might be able to cherry-pick or merge changes from there to complete some not needed constructor-injections.

I'm happy to rebase it on top of this branch too and raise a PR to yours, if you don't want to mess about.

TWiStErRob (approved):
Nice one! I'm curious too, @szpak, if this is a desired changeset to continue working on? Probably the best time for 2.0. I would be happy to help on this.

TWiStErRob:
@3flex re #228, probably temporarily ignore the test that fails and observe green CI without them, just to be sure and then revert.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
3flex: Whoops! Yes, thanks :)

szpak:
> Nice one! I'm curious too, @szpak, if this is a desired changeset to continue working on? Probably the best time for 2.0. I would be happy to help on this.

I have to admit, I don't closely follow the recent changes in Gradle and how the "idiomatic Gradle" looks like in 2023 :-/. However, in general those changes look fine as simplifications and especially with the already mentioned Robert's [branch](https://github.com/TWiStErRob/nexus-publish-gradle-plugin/commits/configuration-cache-induced-idiomatic-gradle-improvements), they would be nicely fit in the upcoming 2.0.0.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
3flex: That's great if you're working on similar changes, I'll leave this one as is, then you can continue on your branch?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -68,7 +82,20 @@ class NexusPublishPlugin : Plugin<Project> {
     }
 
     private fun configureNexusTasks(rootProject: Project, extension: NexusPublishExtension, registry: Provider<InvalidatingStagingRepositoryDescriptorRegistry>) {
+        rootProject.tasks.withType(AbstractNexusStagingRepositoryTask::class.java).configureEach {
```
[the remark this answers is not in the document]
3flex: It just sets defaults. Gradle core plugins are built using the same convention - tasks are preconfigured whenever the plugin is applied and defaults are set.

Using `convention` means it will never override the property's value if the value was `set` somewhere else, so it avoids ordering problems.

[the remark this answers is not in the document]
TWiStErRob: Makes sense. Let's merge then.

szpak (approved):

szpak:
Thanks @3flex for the PR and @TWiStErRob for the code review!

> I think this is good as it is and can be merged, I'm happy to pick up and rebase my branch on top. I can also fix https://github.com/gradle-nexus/publish-plugin/pull/241#discussion_r1255439458 there.

Sound good.

PR #242, opened 2023-07-08
Partially refactor to use more idiomatic Gradle (part 2)

TWiStErRob:
Following https://github.com/gradle-nexus/publish-plugin/pull/241

Recommended to look at the changes commit by commit so the changes make more sense. In the general the goals were:
 * Remove constructor-injection, the Task should be reusable and self-contained, where the data is coming from (i.e. wiring) is the problem of the plugin, not the task.
 * Use the lazy and generated APIs as much as possible.
 * Minor cleanups while doing the above. (e.g. use Kotlin stdlib more, add groups/description for each task)

Breaking changes:
 * `ReleaseNexusStagingRepository.setStagingRepositoryId` renamed to `setStagingRepositoryIdToRelease`
 * `CloseNexusStagingRepository.setStagingRepositoryId` renamed to `setStagingRepositoryIdToClose`
 * All descendants of `AbstractNexusStagingRepositoryTask` have different constructors.


src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -344,7 +365,8 @@ class NexusPublishPlugin : Plugin<Project> {
     }
 }
 
-inline fun <reified T : Any> Project.theExtension(): T =
+private inline fun <reified T : Any> Project.theExtension(): T =
```
TWiStErRob, line 368: I think this is visible in RC1 only, it kind-of pollutes global namespace if someone has similarly named utilities.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -75,80 +79,98 @@ class NexusPublishPlugin : Plugin<Project> {
...
+                this.packageGroup.convention(extension.packageGroup)
+            }
             val findStagingRepository = rootProject.tasks.register<FindStagingRepository>(
                 "find${capitalizedName}StagingRepository",
-                rootProject.objects,
-                extension,
-                repository,
                 registry
             )
             findStagingRepository {
```
TWiStErRob, line 136: My goal was to merge each of these into:
```
val name = rootProject.tasks.register<T>("name") { ... }
```
but these last two I can't figure out how to remove the registry.

I think the answer is using the build service (injecting it inside the task), but that's only possible from Gradle 6.1 (see `NexusPublishPlugin.createRegistry`) and the current minimum is 6.0.

@3flex any ideas?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -75,80 +79,98 @@ class NexusPublishPlugin : Plugin<Project> {
...
         extension.repositories.all {
-            username.convention(rootProject.provider { rootProject.findProperty("${name}Username") as? String })
-            password.convention(rootProject.provider { rootProject.findProperty("${name}Password") as? String })
-            publicationType.convention(PublicationType.MAVEN)
-
             val repository = this
-            val retrieveStagingProfileTask = rootProject.tasks.register<RetrieveStagingProfile>("retrieve${capitalizedName}StagingProfile", rootProject.objects, extension, repository)
+
+            repository.username.convention(rootProject.provider { rootProject.findProperty("${name}Username") as? String })
+            repository.password.convention(rootProject.provider { rootProject.findProperty("${name}Password") as? String })
```
TWiStErRob, line 108: @szpak I wanted to replace this with:
```suggestion
            repository.username.convention(rootProject.gradleProperty("${name}Username"))
            repository.password.convention(rootProject.gradleProperty("${name}Password"))
```
so badly, but that method is only avaiable from Gradle 6.2.

Also it would be an inconvenient breaking change, because right now `findProperty` does a ton of things (see [javadoc](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#findProperty-java.lang.String-)), while `.gradleProperty` is just `-P` on the command line or `gradle.properties`. I think either it needs to be standardized `-P` or `-D`, or use something like:
```kotlin
/**
 * @see Project.findProperty as it was used to find this, but it's impractical do the same in the new lazy API.
 */
private fun ProviderFactory.findProperty(propertyName: String): Provider<String> =
    gradleProperty(propertyName).orElse(systemProperty(propertyName))
```
The rest of the steps in `findProperty` are either deprecated or totally magical.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -75,80 +79,98 @@ class NexusPublishPlugin : Plugin<Project> {
...
-                repository,
                 registry
             )
             findStagingRepository {
-                description = "Finds the staging repository for ${repository.name}"
+                this.group = PublishingPlugin.PUBLISH_TASK_GROUP
+                this.description = "Finds the staging repository for ${repository.name}"
+                this.repository.convention(repository)
+                this.packageGroup.convention(extension.packageGroup)
+                this.descriptionRegex.convention(extension.repositoryDescription.map { "\\b" + Regex.escape(it) + "(\\s|$)" })
```
TWiStErRob, line 141: @szpak These `this.` prefixes might look strange, but without it, future evolutions might break things, because there's a repository (from .all) and a task (from register/invoke) in scope. Alternatively this big setup block could be extracted to a separate function with repository as the parameter not a receiver:
```
        extension.repositories.all {
            extracted(this, rootProject, extension, registry)
        }
```

Edit:I did this in 777dbdeb

TWiStErRob (commented):
Few notes for reviewers.

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
```diff
@@ -16,21 +16,13 @@
...
-import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.options.Option
-import javax.inject.Inject
 
-abstract class CloseNexusStagingRepository @Inject constructor(
-    objects: ObjectFactory,
-    repository: NexusRepository,
-    registry: Provider<InvalidatingStagingRepositoryDescriptorRegistry>
-) : AbstractTransitionNexusStagingRepositoryTask(objects, repository, registry) {
+abstract class CloseNexusStagingRepository : AbstractTransitionNexusStagingRepositoryTask() {
```
TWiStErRob, line 22: I noticed that only abstract tasks are named "...Task", the code might be easier to read if all of them have the suffix, similar to how ...Plugin and ...Extension have suffixes. This would be a bigger breaking change, one that might actually affect users (tasks.withType). Thoughts @szpak?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: Hmm, are you sure that `findProperty` searches also for system properties (`-D`)? Definitely, `-P` was intended to use here (in GNPP).

I wonder, how many people could use the other cases mentioned in javadoc?

Btw, it's misleading - in code review in a browser - to distinguish `orElse(T)` and `orElse(Provider<T>)`. I like a [name separation](https://github.com/gradle/gradle/issues/10520) in Java. However, designing API is hard :-).

Btw2, a fun fact is that I proposed and implemented the initial version of `findProperty` in Gradle years ago (I use so often the tertiary operator in `build.gradle`...). However, the truth is, the aforementioned "magic" it/was itself already implemented at the lower level - https://github.com/gradle/gradle/pull/578/files#diff-2b0a49af16e081b46065d37c7fe53d37b8ee2fc036d9cdbe613f603bfd4e1ef9R778

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishExtension.kt
```diff
@@ -51,15 +50,7 @@ abstract class NexusPublishExtension @Inject constructor(objects: ObjectFactory)
...
-        // `objects.domainObjectContainer(NexusRepository::class) { name -> ... }`,
-        // but in Kotlin 1.3 "New Inference" is not implemented yet, so we have to be explicit.
-        // https://kotlinlang.org/docs/whatsnew14.html#new-more-powerful-type-inference-algorithm
-        objects.domainObjectContainer(
-            NexusRepository::class,
-            NamedDomainObjectFactory { name ->
-                objects.newInstance(NexusRepository::class, name)
-            }
-        )
+        objects.domainObjectContainer(NexusRepository::class)
```
szpak, line 53: Nice. And thanks @3flex for the idea.

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/InvalidatingStagingRepositoryDescriptorRegistry.kt
```diff
@@ -44,11 +44,12 @@ class InvalidatingStagingRepositoryDescriptorRegistry : StagingRepositoryDescrip
...
-                    val invalidateDescriptorMethod = AbstractResolutionAwareArtifactRepository::class.java.declaredMethods.find { it.name.contains("invalidateDescriptor") }
-                    invalidateDescriptorMethod?.isAccessible = true
-                    invalidateDescriptorMethod?.invoke(this)
+                    AbstractResolutionAwareArtifactRepository::class.java
+                        .getDeclaredMethod("invalidateDescriptor")
+                        .apply { isAccessible = true }
+                        .invoke(this)
                 } catch (e: Exception) {
-                    log.warn("Failed to invalidate artifacty repository URL, publishing will not work correctly")
+                    log.warn("Failed to invalidate artifact repository URL, publishing will not work correctly.")
```
szpak, line 52: Good catch...

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: :+1:

szpak (changes requested):
Thanks @TWiStErRob. It looks good, I've just made a few nitpicking.

Btw, as my other plugins are still at the Gradle 6 level of "idiomaticity", @3flex maybe you could also take a look?

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
TWiStErRob: Yeah, definitely separate PR, in IJ Gradle Plugin they did this which is really cool:
https://github.com/JetBrains/gradle-intellij-plugin/blob/89ae190a30b5392ddc36bb02cc4042458c688142/src/main/kotlin/org/jetbrains/intellij/tasks/InstrumentCodeTask.kt#L27C1-L28

But it only works for Kotlin, Java and Groovy users will still get a "missing class" compile error. And any transitive projects will also fail as v1 and v2 will have different classes.

If we do the rename and reintroduce extending classes as you're showing, we would still need to register the old names in the TaskContainer so that `tasks.withType` keeps working. That said, if you want a deprecation cycle, that's the only thing we can do without seriously breaking all users of `withType` (and whatever other use case I'm not thinking of).

Keep in mind that deprecation warnings are not always shown for build.gradle files, so it might be hard for some users to notice.

src/main/kotlin/io/github/gradlenexus/publishplugin/ReleaseNexusStagingRepository.kt
```diff
@@ -16,21 +16,13 @@
...
-abstract class ReleaseNexusStagingRepository @Inject constructor(
-    objects: ObjectFactory,
-    repository: NexusRepository,
-    registry: Provider<InvalidatingStagingRepositoryDescriptorRegistry>
-) : AbstractTransitionNexusStagingRepositoryTask(objects, repository, registry) {
+abstract class ReleaseNexusStagingRepository : AbstractTransitionNexusStagingRepositoryTask() {
 
     @Option(option = "staging-repository-id", description = "staging repository id to release")
-    fun setStagingRepositoryId(stagingRepositoryId: String) {
+    fun setStagingRepositoryIdToRelease(stagingRepositoryId: String) {
```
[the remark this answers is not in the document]
TWiStErRob: moved breaking changes to OP

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
```diff
@@ -16,21 +16,13 @@
...
 
     @Option(option = "staging-repository-id", description = "staging repository id to close")
-    fun setStagingRepositoryId(stagingRepositoryId: String) {
+    fun setStagingRepositoryIdToClose(stagingRepositoryId: String) {
```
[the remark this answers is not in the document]
TWiStErRob: moved breaking changes to OP

TWiStErRob:
> Btw, as my other plugins are still at the Gradle 6 level of "idiomaticity"

Btw, this is motivated by https://github.com/gradle/gradle/pull/22348 and https://docs.google.com/document/d/1iZHN5mcN56IQtGOoR5GyH9NQLSOQLQ1SHN5ulNheDjc which is going towards "managed everything".

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
TWiStErRob: > are you sure that findProperty searches also for system properties (-D)

huh, you're right, that's good then, `providers.gradleProperty(name)` would be enough. Tried `println(findProperty("foo"))`:
 * `gradlew` &rarr; null
 * `gradlew -Pfoo=bar` &rarr; "bar"
 * `gradlew -Dfoo=bar` &rarr; null
 * `gradle.properties` `foo=bar` &rarr; "bar"
 * `gradle.properties` `systemProp.foo=bar` &rarr; null

The limitation still applies: min is Gradle 6.0 now, and the method is only from 6.2.

---

> I wonder, how many people could use the other cases mentioned in javadoc?

The only one I can imagine is (4) `ext.sonatypeUsername` like [this](https://github.com/square/retrofit/commit/4fa58ebb9781502d9afc1ac3c6273b117a667e76#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7), but then the username/password is hardcoded in plain sight. (3) is deprecated, but (2) and (3) could be used to propagate a value from a plugin, like
```
target.extensions.create("sonatypePassword", stringRetrivedFromUnthinkableSource)
```
if that's even valid.

---

> Btw, it's misleading - in code review in a browser - to distinguish orElse(T) and orElse(Provider<T>)

I agree the problem is not with orElse, but with Kotlin's implicit receivers IMO:
```kotlin
private fun ProviderFactory.findProperty(propertyName: String): Provider<String> =
    this.gradleProperty(propertyName).orElse(this.systemProperty(propertyName))
```
in this case it's clear(er) that both of them are providers. and that `systemProperty` is not a theoretical `project.systemProperty(name): String`.

---

> Btw2, a fun fact is that I proposed and implemented the initial version of findProperty

Haha, nice! 👏 reminds me: "designing API is hard" ... `project.propertyOrNull(name)` would be the name according to Kotlin stdlib conventions, which I think are really convenient.

---

Btw3, when I first came to this repo, I knew your name was familiar from somewhere... know I know, it's [PIT](https://github.com/szpak/gradle-pitest-plugin)!

TWiStErRob:
@szpak Ok, so in the end there were no changes necessary, the only finding you flagged was in uncommitted code 😅.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: > The limitation still applies: min is Gradle 6.0 now, and the method is only from 6.2.

I see no problem. People using 6.0 should be able to upgrade to 6.2 without any issues, so let's bump the minimal Gradle version to 6.2.

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
szpak: > Keep in mind that deprecation warnings are not always shown for build.gradle files, so it might be hard for some users to notice.

You are right, we would need to do some warning logging before delegation or use some internal Gradle magic for presenting deprecation warnings. To limit (that) plugin maintenance cost, I'm open to just rename the classes - 2.0.0 might provide some surprises :-).

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
TWiStErRob: Let's leave it as is for this PR, I might open another one to see what a 6.2 bump would look like.

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
TWiStErRob: Alternatively, we can just leave it as is :)

Anyway, let's see how it would look like in a separate PR and continue discussion there.

TWiStErRob:
I consider this ready, pending a word from @3flex.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: Sound good.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
TWiStErRob: Just moved from unannotated constructor property to @Internal managed property: https://github.com/gradle-nexus/publish-plugin/pull/242/commits/e964697bcaf1e996feea3fc319df06777445bba2

src/main/kotlin/io/github/gradlenexus/publishplugin/CloseNexusStagingRepository.kt
TWiStErRob: Opened https://github.com/TWiStErRob/nexus-publish-gradle-plugin/pull/6 for now until this gets merged.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
TWiStErRob: Opened https://github.com/TWiStErRob/nexus-publish-gradle-plugin/pull/7 for now until this PR gets merged.

szpak (approved):

TWiStErRob:
Ugh, should I rebase and squash a few next time to not get so many random small commits to master?

szpak:
> Ugh, should I rebase and squash a few next time to not get so many random small commits to master?

No, I did it purposely. I wanted to keep some of those smaller changes (to maybe replicate in my other plugins ;) ). Sure, some commits could be squashed/rebased, but I didn't want to bother you.
In general, a "squash and commit" is a default strategy set for the project.

PR #250, opened 2023-07-18
Breaking change: rename Tasks to have Task suffix

TWiStErRob:
Planned 2.0 breaking change from https://github.com/gradle-nexus/publish-plugin/pull/242#discussion_r1257343249

E2E tests will fail because each has a reference to old names:
```
tasks.withType(io.github.gradlenexus.publishplugin.InitializeNexusStagingRepository).configureEach {
```

---

Raised as a draft, because I'm not fully sure this is worth the maintenance cost in the community (not much on GitHub, but we don't know the full scale of closed source):
 * https://github.com/search?q=%2F%28%3F-i%29%5CbRetrieveStagingProfile%5Cb%2F&type=code
 * https://github.com/search?q=%2F%28%3F-i%29%5CbFindStagingRepository%5Cb%2F&type=code
 * https://github.com/search?q=%2F%28%3F-i%29%5CbInitializeNexusStagingRepository%5Cb%2F&type=code
 * https://github.com/search?q=%2F%28%3F-i%29%5CbReleaseNexusStagingRepository%5Cb%2F&type=code
 * https://github.com/search?q=%2F%28%3F-i%29%5CbCloseNexusStagingRepository%5Cb%2F&type=code

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,8 +201,8 @@ By default, `initialize${repository.name.capitalize()}StagingRepository` task ad
 
 The description can be customized via:
 * `io.github.gradlenexus.publishplugin.NexusPublishExtension.getRepositoryDescription` property (default: `$group:$module:$version` of the root project)
-* `io.github.gradlenexus.publishplugin.InitializeNexusStagingRepository.repositoryDescription` property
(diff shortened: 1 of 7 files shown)


szpak:
We discussed it with @TWiStErRob and the benefits in code maintenance don't seem to overweight the breaking builds that use the tasks directly (and even more, the build broken due to the plugins that use those tasks directly). So, closing it for now.

PR #245, opened 2023-07-13
Random fixes

TWiStErRob:

PR #251, opened 2023-07-18
Gradle 6.2 as minimum

TWiStErRob:
As promised, based on https://github.com/gradle-nexus/publish-plugin/pull/242#discussion_r1257338630

Doesn't give too many improvements, it actually looks more complicated 😅.


szpak:
To be honest, I'm not sure about that chance. I'm ambivalent to it. What do you propose?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -100,13 +98,14 @@ class NexusPublishPlugin : Plugin<Project> {
...
             transitionCheckOptions.convention(extension.transitionCheckOptions)
-            stagingRepositoryId.convention(registry.map { it[repository.get().name].stagingRepositoryId })
+            usesService(registryService)
+            stagingRepositoryId.convention(registryService.map { it.registry[repository.get().name].stagingRepositoryId })
         }
         extension.repositories.all {
-            username.convention(rootProject.provider { rootProject.findProperty("${name}Username") as? String })
-            password.convention(rootProject.provider { rootProject.findProperty("${name}Password") as? String })
+            username.convention(rootProject.providers.gradleProperty("${name}Username").forUseAtConfigurationTimeCompat())
+            password.convention(rootProject.providers.gradleProperty("${name}Password").forUseAtConfigurationTimeCompat())
```
TWiStErRob, line 106: Behavior change as discussed in https://github.com/gradle-nexus/publish-plugin/pull/242#discussion_r1257338630

> To be honest, I'm not sure about that chance. I'm ambivalent to it. What do you propose? - @szpak 

Assuming you're referring to the chance of someone using something other than `-P`: I think a major version change is the best time to introduce breaking changes like this. It won't come unexpected. And we can list it in a migration guide as something to look out for.

TWiStErRob: ... and if someone does use it, and want it, they'll hopefully tell us :)

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -54,8 +53,8 @@ class NexusPublishPlugin : Plugin<Project> {
             "Plugin must be applied to the root project but was applied to ${project.path}"
         }
 
-        require(GradleVersion.current() >= GradleVersion.version("6.0")) {
-            "The plugin requires Gradle version 6.0+"
+        require(GradleVersion.current() >= GradleVersion.version("6.2")) {
+            "The plugin requires Gradle version 6.2+"
```
TWiStErRob, line 57:
```suggestion
            "`io.github.gradle-nexus.publish-plugin` requires Gradle version 6.2+"
```

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: I was rather writing about benefits of bumping required version to 6.2 :-). Nevertheless, the newer API we use, the (hopefully) later it will be deprecated in Gradle :-).

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -54,8 +53,8 @@ class NexusPublishPlugin : Plugin<Project> {
             "Plugin must be applied to the root project but was applied to ${project.path}"
         }
 
-        require(GradleVersion.current() >= GradleVersion.version("6.0")) {
-            "The plugin requires Gradle version 6.0+"
+        require(GradleVersion.current() >= GradleVersion.version("6.2")) {
```
TWiStErRob, line 56: If only we had some stats of Gradle version distributions out in the wild, like Android does: https://www.androidauthority.com/wp-content/uploads/2023/06/Android-version-distribution-statistics-May-2023-988w-675h.jpg.webp

It would be a much easier to say "minimum x".

szpak: I asked about that one friendly Gradle engineer ;-). We will see.

szpak:
I feel convinced. Bumping from 6.0 to 6.2 should be easy for the majority of projects. Could you resolve the conflicts?

TWiStErRob:
Will do later.

szpak (approved):

szpak:
I'm afraid, we hit a regression in e2e tests related to the properties handling:
https://scans.gradle.com/s/4b3c4szjffcbg/tests/overview?outcome=failed

> Caused by: java.lang.RuntimeException: Failed to create staging repository, server at https://oss.sonatype.org/service/local/ responded with status code 401, body:

However, as we do not check authorization (I've just created #256 as possible better regression testing passing credentials to Nexus in general), I'm not sure, if it is a problem with "gradleProperty" or we stopped passing credentials to Nexus at all (rather not).

I plan to debug the e2e tests tomorrow to find out.

TWiStErRob:
Could the regression be this one from another PR?
https://github.com/gradle-nexus/publish-plugin/pull/245/commits/e37d8476fc76f1009fd40d608d96a649fe92cfc5

szpak:
> Could the regression be this one from another PR? [e37d847](https://github.com/gradle-nexus/publish-plugin/commit/e37d8476fc76f1009fd40d608d96a649fe92cfc5)

Rather not, e2e tests passed after that commit: https://github.com/gradle-nexus/publish-plugin/actions/runs/5561323077

It started failing after this MR was merged:
https://github.com/gradle-nexus/publish-plugin/actions?query=workflow%3A%22E2E+tests%22

szpak:
Btw, just in case you would like to test some "risky" change with the e2e tests, it is enough to have `e2e/` branch prefix (and push the branch directly into that repo, not the fork):
https://github.com/gradle-nexus/publish-plugin/blob/master/.github/workflows/e2e.yml#L9

szpak:
You can also run the workflow manually for any branch in that repo, using the button on the right:
https://github.com/gradle-nexus/publish-plugin/actions/workflows/e2e.yml

image

(I don't know, if it is possible to configure it to execute manually also for - safe - pull requests)

PR #254, opened 2023-07-22
Build on latest Java, test through lowest possible

TWiStErRob:
Includes #247 change.


build.gradle.kts
```diff
@@ -260,6 +258,11 @@ tasks {
     withType<Test>().matching { it.name.startsWith("compatTest") }.configureEach {
         systemProperty("plugin.version", project.version)
     }
+    named<Test>("test").configure {
+        javaLauncher = project.javaToolchains.launcherFor {
```
TWiStErRob, line 262: Note: End to end tests are left on Gradle 11, however this could mean we're testing something that the users will never see! (Btw, where is the publish workflow or release steps?)

For E2E We should probably
 * set Gradle to 19
 * leave the compilation as default (i.e. 19)
 * set the runner toolchain to something explicit
    * probably a matrix of 8, 11, 17!

TWiStErRob: https://github.com/gradle-nexus/publish-plugin/pull/254/commits/b2dff42a5bbb6f1d9bf3468392b1c2c3e5381da5

szpak (changes requested):

.github/workflows/e2e.yml
```diff
@@ -11,7 +11,13 @@ on:
 
 jobs:
   tests:
+    name: "E2E tests on Java ${{ matrix.jdk }}"
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        jdk: [ 8, 11, 17 ]
```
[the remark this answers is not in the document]
TWiStErRob: To reduce the load, we can reduce the matrix for `push`, but I think at least right before release it should be executed on all versions via workflow_dispatch. (Conditional CI matrix is possible without too much hassle.)

compatTests are by definition separated by Java versions, the task name contains it.
foojay-resolver does the installing on dev machines as well as CI, so it's not a big deal to have any version. That said I added `auto` for devs just to make it more native to the machine.

PR #258, opened 2023-07-25
Proof of concept: Gradle 8.3

TWiStErRob:
This will allow us to use Gradle 8.3+, but we'll have to live with the warning:

> Configure project :
This version of Gradle expects version '4.1.0' of the `kotlin-dsl` plugin but version '4.0.14' has been applied to root project 'publish-plugin'. Let Gradle control the version of `kotlin-dsl` by removing any explicit `kotlin-dsl` version constraints from your build logic.
WARNING: Unsupported Kotlin plugin version.
The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `1.9.0` that might work differently than in the requested version `1.8.22`.

### DIFF
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/internal/StagingRepositoryTransitioner.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/internal/StagingRepositoryTransitioner.kt
@@ -59,7 +59,7 @@ class StagingRepositoryTransitioner(val nexusClient: NexusClient, val retrier: A
     private fun assertRepositoryInDesiredState(repository: StagingRepository, vararg desiredStates: StagingRepository.State) {
         if (repository.state !in desiredStates) {
             throw RepositoryTransitionException(
-                "Staging repository is not in desired state ${desiredStates.contentToString()}: $repository. It is unexpected. Please check " +
+                "Staging repository is not in desired state ${desiredStates.contentDeepToString()}: $repository. It is unexpected. Please check " +
                     "the Nexus logs using its web interface - it can be caused by validation rules violation (e.g. publishing artifacts with the " +
                     "same version again). If not, please report it to https://github.com/gradle-nexus/publish-plugin/issues/ with the '--info' logs."
             )
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,10 +5,21 @@ import org.gradle.initialization.IGradlePropertiesLoader.SYSTEM_PROJECT_PROPERTI
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 
+buildscript {
+    configurations.classpath.get().dependencyConstraints.apply {
+        removeAll(filter { it.reason == "Pinned to the embedded Kotlin" }.toSet())
+    }
+    dependencies {
+        classpath(enforcedPlatform("org.jetbrains.kotlin:kotlin-bom:1.8.22"))
+    }
+}
+
 plugins {
     // Leave this version behind of the current Gradle version, on the level of Gradle 8.2.1.
     // See https://github.com/gradle/gradle/issues/25868 for more details.
     `kotlin-dsl` version "4.0.14"
+    // Last version of Kotlin Gradle Plugin to support Kotlin 1.3 back-compilation.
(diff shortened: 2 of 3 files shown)


src/main/kotlin/io/github/gradlenexus/publishplugin/internal/StagingRepositoryTransitioner.kt
```diff
@@ -59,7 +59,7 @@ class StagingRepositoryTransitioner(val nexusClient: NexusClient, val retrier: A
     private fun assertRepositoryInDesiredState(repository: StagingRepository, vararg desiredStates: StagingRepository.State) {
         if (repository.state !in desiredStates) {
             throw RepositoryTransitionException(
-                "Staging repository is not in desired state ${desiredStates.contentToString()}: $repository. It is unexpected. Please check " +
+                "Staging repository is not in desired state ${desiredStates.contentDeepToString()}: $repository. It is unexpected. Please check " +
```
szpak, line 62: Is this a backward compatible change? When was it introduced (before 6.2?)?

szpak (approved):
Btw, as I see in our [regression tests](https://github.com/gradle-nexus/publish-plugin/actions/runs/5641947708/job/15280868434#step:6:14) the plugin can already be built with 8.3-rc-1 and the change it to "just" make the project itself "buildable" with it?

If you think, the changes are safe (and backward compatible at runtime), it's ok for me to apply them to be able to keep up with the changes in Gradle even more :).

Btw. what will be the minimal Gradle version able to build the project? 8.0?

src/main/kotlin/io/github/gradlenexus/publishplugin/internal/StagingRepositoryTransitioner.kt
TWiStErRob: [`contentToString`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/content-to-string.html) seems actually Kotlin 1.4, and we're targeting Kotlin 1.3. Not sure how it was compiling before. [`contentDeepToString`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/content-deep-to-string.html) is Kotlin 1.1

It's very likely some kind of inference / overload resolution change.

TWiStErRob:
> If you think, the changes are safe (and backward compatible at runtime), it's ok for me to apply them to be able to keep up with the changes in Gradle even more :).

Sorry this was meant to be a draft only, for future reference. For now this is RC, let's at least wait for stable. Things can change significantly in RCs (if they are broken in RC1, they might be fixed in RC2).

This PR might not even be necessary if we bump the minimum Gradle to 7.x, but that's a different convo. Let's get this 2.0 out first :)

> Btw, as I see in our [regression tests](https://github.com/gradle-nexus/publish-plugin/actions/runs/5641947708/job/15280868434#step:6:14) the plugin can already be built with 8.3-rc-1 and the change it to "just" make the project itself "buildable" with it?

The regression tests are running on Gradle 8.3-rc-1, still using the wrapper to build.
https://github.com/gradle-nexus/publish-plugin/actions/runs/5641947708/job/15280868434#step:7:91
image
So this change updates the wrapper for all `gradlew` commands.

> Btw. what will be the minimal Gradle version able to build the project? 8.0?

Not sure I get this question. Wrapper locks in a specific version. Using the plugin (applying it) in a (user) build depends on what we set explicitly as the minimum in the plugin code.

szpak:
> The regression tests are running on Gradle 8.3-rc-1, still using the wrapper to build.

Yes, I meant, that the currently built plugin (with 8.2) is at runtime compatible with (in) the 3rd-party projects built with 8.3-rc-1.

> Not sure I get this question. Wrapper locks in a specific version. Using the plugin (applying it) in a (user) build depends on what we set explicitly as the minimum in the plugin code.

Yes, the wrapper keeps the things consistent. I was just wondering, if - due to that changes - the "builability" with the lower Gradle version is impossible (however, it was just a theoretical question, as 8.2+ should be enough for us).

szpak:
And yes, waiting for 8.3-final seems to be a good idea :-).

TWiStErRob:
I think 8.5 will be even harder, hopefully I can have a look this weekend. I'll try to carve a few hours for reviving these PRs. But please review closely after, because it's really hard to remember the context and rebase these with confidence.

Issue #220, opened 2023-05-07
`repositories` function clashes with `Project.repositories` from the Kotlin DSL (Gradle 8.1)


szpak:
Unrelated to the above but also related to Gradle 8 compatibility (at least with Gradle 8.1.1):

![image](https://user-images.githubusercontent.com/8314616/234118093-edb8f874-60e7-450f-9b60-7796629ae494.png)

The `repositories` function is now clashing with `Project.repositories` from the Kotlin DSL. For now it seems to still compile, but it does show an error in the IDE. Doing something like the following fixes it (since thankfully the `Project.repositories` argument has a different name):

![image](https://user-images.githubusercontent.com/8314616/234118587-bf553f01-537d-409a-94d1-f57ed7d34aa8.png)

It might make sense to introduce an alias to `repositories` such as `nexusRepositories` and perhaps deprecate the former.

If I'm missing something obvious and there's a less hacky workaround please do let me know!

I tried going over the Gradle release notes to check if this change in behaviour is mentioned anywhere but couldn't find anything to reference. I'd appreciate if someone could find a reference as to why this is now happening.

Also, please let me know if you'd prefer that I create a new issue to track this.

_Originally posted by @YarnSphere in https://github.com/gradle-nexus/publish-plugin/issues/208#issuecomment-1520845369_

szpak:
@YarnSphere seems to [find a workaround](https://github.com/gradle-nexus/publish-plugin/issues/208#issuecomment-1529047972):

---

Regarding https://github.com/gradle-nexus/publish-plugin/issues/208#issuecomment-1520845369, where I say:

> If I'm missing something obvious and there's a less hacky workaround please do let me know!

The less hacky workaround is simply:

```kt
nexusPublishing {
    this.repositories {
        sonatype()
    }
}
```

Which I find acceptable (I don't know how I originally missed this :sweat_smile:).

---

I wonder if - in addition to mention it in the documentation - the plugin should do anything else (considering the backward compatibility)?

TWiStErRob:
> For now it seems to still compile, but it does show an error in the IDE.

This means the regression bug is in the IDE or the Gradle v IDEA integration. Is there a repro that we can use to report it to Gradle?

Vampire:
Seems to work just fine with Gradle 8.2 and IntelliJ IDEA 2023.1.3:
[...]

PR #260, opened 2023-07-31
Fixing IDE errors

DenuxPlays:
Fixes: #220 (or at least related to)


MoonTM-GIT (approved):

marcphilipp:
@DenuxPlays Thanks, but I think we should instead wait for the fix in IDEA (see https://github.com/gradle-nexus/publish-plugin/issues/220#issuecomment-1697114625)

Issue #236, opened 2023-07-04
Kotlin DSL friendly overall tasks


Vampire:
You have the `closeAndReleaseStagingRepository` task (which should probably better be called `closeAndReleaseStagingRepositories`).
This task is always registered no matter which repositories are configured.
This is good, as then we also get type-safe accessors for the Kotlin DSL for it.
And as typically only Maven Central is configured, this is also a perfect replacement for the `closeAndReleaseSonatypeStagingRepository` task that does not get accessors generated.

The problem is, that I don't want to do close and release coupled.
I do the closing when doing also other publishing tasks like creating the GitHub publication.
Only after also all other tasks were successful, I want to release the repository.

So I would welcome `closeStagingRepositor(y/ies)` and `releaseStagingRepositor(y/ies)` to conveniently wire them using Kotlin DSL.

sschuberth:
Looks like this issue is actually not so much about Kotlin DSL, but somewhat duplicates https://github.com/gradle-nexus/publish-plugin/issues/215 and https://github.com/gradle-nexus/publish-plugin/issues/134 (and is quite the opposite of https://github.com/gradle-nexus/publish-plugin/issues/77).

Vampire:
Not really @sschuberth.
This is about adding summary tasks for only closing and only releasing like there is one for closing and releasing together.
This is not too tightly related to Kotlin DSL, but makes the usage better in Kotlin DSL as described above.

#215 is about the standalone closing and releasing somehow having some problem
#134 is about the possibility to upload multiple things in multiple Gradle runs and then do a close and release for them together in a separate run
#77 is not at all opposite to this request, it just requests some combination tasks that that combine publishing and closing which is totally independent of this request

sschuberth:
> Well, while we are at it, I implemented it in #285 :-D

[Good luck](https://github.com/gradle-nexus/publish-plugin/issues/156) with getting it merged (honestly, no sarcasm here).

PR #285, opened 2023-11-23
Add closeStagingRepository and releaseStagingRepository summary tasks (#236)

Vampire:


src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
```diff
@@ -106,6 +106,32 @@ class TaskOrchestrationTest {
...
+        initSingleProjectWithDefaultConfiguration()
+        project.extensions.configure<NexusPublishExtension> {
+            repositories.clear()
+        }
+
+        val simplifiedReleaseTasks = project.getTasksByName(NexusPublishPlugin.SIMPLIFIED_RELEASE_TASK_NAME, true)
+
+        assertThat(simplifiedReleaseTasks).hasSize(1)
+        assertThat(simplifiedReleaseTasks.first().dependsOn).isEmpty()
+    }
```
szpak, line 133: What do you think about using the parameterized tests for those two sibling cases?

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
```diff
@@ -135,6 +191,34 @@ class TaskOrchestrationTest {
...
+        project.extensions.configure<NexusPublishExtension> {
+            repositories.create("otherNexus")
+        }
+
+        val releaseTask = getJustOneTaskByNameOrFail(NexusPublishPlugin.SIMPLIFIED_RELEASE_TASK_NAME)
+
+        assertThat(releaseTask.description)
+            .contains("sonatype")
+            .contains("otherNexus")
+    }
```
szpak, line 220: The same here.

szpak (changes requested):
Thanks for your contribution.

I agree that they might be useful. I believe, there was a similar discussion years ago and (AFAIR) the main argument against, given by someone, was increased number of tasks (to quickly understand what should be used for new users) and a chance to (try to) accidentally close "too many" repositories.

While the first one might be some problem for newcomers, more experiences users could be happy having them. In the second case, I think that having 2+ repositories configured requires some "knowledge" about Gradle and people should be aware of the pitfalls.

Wrapping up, I'm ok for that contribution as it is more a matter of taste. @TWiStErRob How do you see it?

Btw, @Vampire it would be good to have them mentioned in README to make people more aware of that "simplification".

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
Vampire: Sure, can you point me to an example maybe? I'm not too much into Jupiter, I'm still largely preferring Spock. :-)

szpak: :-D

https://github.com/gradle-nexus/publish-plugin/blob/00d5611cfdee82ecd2e26eeef1aac40e61a9a246/src/e2eTest/kotlin/io/github/gradlenexus/publishplugin/e2e/NexusPublishE2ETests.kt#L29-L31

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
szpak: In that case with more parameter, `@CsvSource` might be more useful, especially the new `textBlock` parameter which tries to mimic a table from Spock:
https://junit.org/junit5/docs/current/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/CsvSource.html#textBlock()

Vampire: Sorry, seen too late, I now used a `@MethodSource` but I can change it if you want.
Anyway, I'm now reminded of one point why I prefer Spock. :-D

README.md
```diff
@@ -211,6 +211,26 @@ So the steps to publish and release in different Gradle invocations are:
...
+- `closeOtherNexusStagingRepository`
+- `releaseSonatypeStagingRepository`
+- `releaseOtherNexusStagingRepository`
+- `closeAndReleaseSonatypeStagingRepository`
+- `closeAndReleaseOtherNexusStagingRepository`
+
+For convenience there are also summary tasks generated, that group the tasks for the different repositories, which are
+- `closeStagingRepository`
+- `releaseStagingRepository`
+- `closeAndReleaseStagingRepository`
```
TWiStErRob, line 228: since master is at a 2.x stage, why not go with your original observation/idea, @Vampire?
```suggestion
- `closeStagingRepositories`
- `releaseStagingRepositories`
- `closeAndReleaseStagingRepositories`
```

@szpak acceptable?

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
```diff
@@ -93,44 +96,77 @@ class TaskOrchestrationTest {
...
             .hasSize(2)
-            .contains("closeAndReleaseSonatypeStagingRepository")
-            .contains("closeAndReleaseOtherNexusStagingRepository")
+            .contains(sonatypeTaskName)
+            .contains(otherNexusTaskName)
     }
 
-    @Test
-    internal fun `description of simplified closeAndRelease task contains names of all defined Nexus instances`() {
+    private fun `simplified task without repository name should depend on all normal tasks (created one per defined repository) arguments`(): Stream<Arguments> = Stream.of(
```
TWiStErRob, line 135: I would go for CsvSource, this seems a ton of code for what it does.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -125,10 +127,17 @@ class NexusPublishPlugin : Plugin<Project> {
                 enabled = false
             }
         }
-        rootProject.tasks.register<Task>(SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME) {
+        rootProject.tasks.register(SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME) {
             group = PublishingPlugin.PUBLISH_TASK_GROUP
             description = "Closes and releases open staging repositories in all configured Nexus instances."
-            enabled = false
```
TWiStErRob, line 131: What was this `enabled = false/true` about? Is it safe to remove, or we need it for the other ones too?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -47,6 +47,8 @@ class NexusPublishPlugin : Plugin<Project> {
     companion object {
         // visibility for testing
         const val SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME = "closeAndReleaseStagingRepository"
+        const val SIMPLIFIED_CLOSE_TASK_NAME = "closeStagingRepository"
+        const val SIMPLIFIED_RELEASE_TASK_NAME = "releaseStagingRepository"
```
TWiStErRob, line 51:
```suggestion
        private const val SIMPLIFIED_CLOSE_AND_RELEASE_TASK_NAME = "closeAndReleaseStagingRepository"
        private const val SIMPLIFIED_CLOSE_TASK_NAME = "closeStagingRepository"
        private const val SIMPLIFIED_RELEASE_TASK_NAME = "releaseStagingRepository"
```

There's no harm repeating the task names in test annotations! It actually gives more hardening quality to the tests, i.e. if someone changes the constants, tests will break, which is a good thing as they're public API.

TWiStErRob (commented):
I like lifecycle/convenience tasks if they're well documented, which seems to be the case (descriptions and groups are clear, naming could be improved).

README.md
Vampire: My guess is, that the main use-case is publishing to Maven Central so 95 % of the users will only have one repository. But technically I agree with "...ies" - of course, it was me mourning about it :-D.
I just made the `...y` for the two new ones to be consistent with the existing one and not sneak in a breaking change.

But I follow what you as maintainers want, just tell me how you like it. :-)

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
Vampire: It is a task without actions, it never does anything anyway. Disabling or enabling it is pointless.

README.md
TWiStErRob: Interestingly using plural might lead more users to realizing that it's possible to do multiple, and that could lead to them having "snapshot" and "mavenCentral" repositories, but this might be a stretch.

If we go ahead consider `closeAllStagingRepositories` naming as IMO that makes the plural thing valid, even if there's only one.

For the new ones `...y` makes sense, as it always only touches 1 repo. But the new ones, make the old ones stand out more with them not being plural. Your docs really highlight it :)

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
TWiStErRob: Could enabling it yield to some more tasks executed (via dependsOn) and therefore breaking some builds which just do blanket `gradlew closeAndReleaseStagingRepository` in a multimodule build?

TWiStErRob: Was it maybe disabled so that `gradlew tasks` doesn't list it when there are no repositories registered?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
Vampire: Sure, I just followed existing style, fixed now.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
Vampire: A no to both, but you are welcome to come up with other options I might have forgotten. :-)

Setting a task to `enabled = false` does not prevent its dependencies to run.
They are still run, just the task action itself is skipped and there is none here.
To skip a task and its dependencies (if nothing else depends on them) you have to use `-x theTask` or change `gradle.startParameter.excludedTaskNames` which is the same.
The `enabled` property on a task without actions is really just a no-op.

There is only one exception actually now that I think of it.
If someone got the idea to register additional actions on the task using `doLast` or `doFirst` then it would make a difference.
But in reality the tasks are never disabled anyway, because why should someone apply the plugin and then not define any repositories. :-)

> Was it maybe disabled so that gradlew tasks doesn't list it when there are no repositories registered?

Not in any Gradle version I am aware of, and that is 1.0-milestone-4 and newer iirc. :-)
```plain
Publishing tasks
----------------
closeAndReleaseStagingRepository - Closes and releases open staging repositories in all configured Nexus instances.
```
You see from the description that it was not adjusted for listing the repositories and this is with the latest published version and only applying it without configuration.

src/test/kotlin/io/github/gradlenexus/publishplugin/TaskOrchestrationTest.kt
Vampire: fixed

README.md
Vampire: > that could lead to them having "snapshot" and "mavenCentral" repositories, but this might be a stretch.

That would be strange actually imho. Why should one have those both in parallel? You never want a version published to both. You even cannot by version checks of Sonatype iirc.

> If we go ahead consider closeAllStagingRepositories naming as IMO that makes the plural thing valid, even if there's only one.

Imho the "All" does not make it any more valid, neither with only one nor with multiple.
But I don't have strong feelings about that, just tell me what you want. :-)

> For the new ones ...y makes sense, as it always only touches 1 repo.

I'm not sure what you call "new ones".
The new ones introduced here are `closeStagingRepository` and `releaseStagingRepository`, which are exactly the ones that could touch multiple repositories and thus "...y" does not actually makes sense.

> But the new ones, make the old ones stand out more with them not being plural. 

I'm afraid I also did not fully get this sentence. :-/

szpak: While technically one could have 2+ (release) repositories configured, I also think simultaneously publishing can be more tricky as the operations are not transactional and the most popular (in the context of two repositories) is releasing to Gradle Plugin Portal and Maven Central (where the former is using the other publishing "technology").

I would expect to use `closeStagingRepository` and `releaseStagingRepository` in 95%+ as a shortcut for releasing to just one place. Nevertheless, I like to idea to emphasis for users that actually there could be more and - in the end - I think that "-ies" is more (logically) accurate. It seems that we all like "-ies" and 2.x seems to be a good time to do it.

Vampire: Ok, changed

szpak (approved):

szpak:
Thanks @Vampire (also for using the `textBlock`). Rebased with master and force pushed.

README.md
TWiStErRob: My bad, I probably messed up "old" and "new", anyway, you got the points :) I shouldn't do reviews that late in the evening.

You're right that there's no point publishing a stable and a snapshot at the same time, but I like szpak's point about GPP and MC, even though it's not actually thing (yet?). I have a feeling in the corporate worlds there are some strange setups where they actually publish to multiple repos.

TWiStErRob (approved):
Thanks for your patience, sorry for the wait.

I think it ended being nice and clean. Looks reasonable to me.

These's one final nit: the title of the PR (and therefore the commit on `master`) needs `-ies`, right?

Vampire:
Absolutely, thanks.
Fixed both, commit message and PR title

szpak:
Thanks again, merged.

Btw, using the merge queue makes it problematic to change the squash commit message :-/

PR #274, opened 2023-09-18
Gradle 6.2 as minimum (re-merge of #251)

TWiStErRob:
* Fixes #267
 * Re-merge of #251
 * 251 was reverted in #268
 * This PR reverts #268

PR #295, opened 2024-01-14
Drop kotlin-dsl plugin

3flex:
Using the `kotlin-dsl` plugin creates additional implicit dependencies on the Kotlin version used by Gradle as seen in https://github.com/gradle-nexus/publish-plugin/pull/258#issue-1820712729.

Removing the use of kotlin-dsl in the plugin itself removes that dependency. Now the Kotlin version required by the project can be set independently and is not tied to the Kotlin version used by the Gradle version applied to the wrapper.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,7 +6,7 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 
 plugins {
-    kotlin("jvm") version "1.8.22"
+    id("org.jetbrains.kotlin.jvm") version "1.8.22"
     id("com.gradle.plugin-publish") version "1.2.1"
     id("com.diffplug.spotless") version "6.23.3"
     id("com.github.johnrengelman.shadow") version "8.1.1"
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,7 +6,7 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 
 plugins {
-    `kotlin-dsl`
+    kotlin("jvm") version "1.8.22"
     id("com.gradle.plugin-publish") version "1.2.1"
     id("com.diffplug.spotless") version "6.23.3"
     id("com.github.johnrengelman.shadow") version "8.1.1"
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -95,6 +95,8 @@ dependencies {
     // but we are running CI on Java 8 in .github/workflows/java-versions.yml.
     testImplementation("org.mockito:mockito-junit-jupiter:4.11.0")
     testImplementation("com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0")
+
+    testImplementation(gradleApi())
 }
 
 java {
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
@@ -223,6 +223,7 @@ class NexusPublishPlugin : Plugin<Project> {
                         val id = when (publicationType) {
                             PublicationType.IVY -> "ivy-publish"
                             PublicationType.MAVEN -> "maven-publish"
+                            null -> error("Repo publication type must be \"ivy-publish\" or \"maven-publish\"")
                         }
                         publishingProject.plugins.withId(id) {
                             val initializeTask = rootProject.tasks.named("initialize${nexusRepo.capitalizedName}StagingRepository", InitializeNexusStagingRepository::class.java)
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
@@ -170,7 +170,7 @@ class NexusPublishPlugin : Plugin<Project> {
             it.usesService(registryService)
             it.repository.convention(repo)
             it.packageGroup.convention(extension.packageGroup)
-            it.descriptionRegex.convention(extension.repositoryDescription.map { "\\b" + Regex.escape(it) + "(\\s|$)" })
+            it.descriptionRegex.convention(extension.repositoryDescription.map { repoDescription -> "\\b" + Regex.escape(repoDescription) + "(\\s|$)" })
         }
         val closeTask = tasks.register(
             "close${repo.capitalizedName}StagingRepository",
@@ -230,9 +230,9 @@ class NexusPublishPlugin : Plugin<Project> {
                             val findStagingRepositoryTask = rootProject.tasks.named("find${nexusRepo.capitalizedName}StagingRepository", FindStagingRepository::class.java)
                             val closeTask = rootProject.tasks.named("close${nexusRepo.capitalizedName}StagingRepository", CloseNexusStagingRepository::class.java)
                             val releaseTask = rootProject.tasks.named("release${nexusRepo.capitalizedName}StagingRepository", ReleaseNexusStagingRepository::class.java)
-                            val publishAllTask = publishingProject.tasks.register("publishTo${nexusRepo.capitalizedName}") {
-                                it.group = PublishingPlugin.PUBLISH_TASK_GROUP
-                                it.description = "Publishes all Maven/Ivy publications produced by this project to the '${nexusRepo.name}' Nexus repository."
+                            val publishAllTask = publishingProject.tasks.register("publishTo${nexusRepo.capitalizedName}") { task ->
+                                task.group = PublishingPlugin.PUBLISH_TASK_GROUP
+                                task.description = "Publishes all Maven/Ivy publications produced by this project to the '${nexusRepo.name}' Nexus repository."
                             }
                             closeTask.configure { task ->
                                 task.mustRunAfter(publishAllTask)
@@ -315,9 +315,9 @@ class NexusPublishPlugin : Plugin<Project> {
             publishTask.configure {
                 it.dependsOn(initializeTask)
                 it.mustRunAfter(findStagingRepositoryTask)
-                it.doFirst {
+                it.doFirst { task ->
                     if (artifactRepo is UrlArtifactRepository) {
-                        it.logger.info("Uploading to {}", artifactRepo.url)
+                        task.logger.info("Uploading to {}", artifactRepo.url)
                     }
                 }
             }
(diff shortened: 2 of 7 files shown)


src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -28,15 +27,9 @@ import org.gradle.api.artifacts.repositories.UrlArtifactRepository
...
+import org.gradle.api.reflect.TypeOf.typeOf
 import org.gradle.api.tasks.TaskContainer
 import org.gradle.api.tasks.TaskProvider
-import org.gradle.kotlin.dsl.create
-import org.gradle.kotlin.dsl.invoke
-import org.gradle.kotlin.dsl.named
-import org.gradle.kotlin.dsl.register
-import org.gradle.kotlin.dsl.registerIfAbsent
-import org.gradle.kotlin.dsl.typeOf
-import org.gradle.kotlin.dsl.withType
```
TWiStErRob, line 39: We can keep these by adding a dependency on kotlin-dsl (not the plugin). It's safe because they're mostly inline afaik and the artifact doesn't change much.

build.gradle.kts
```diff
@@ -6,7 +6,7 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 
 plugins {
-    `kotlin-dsl`
+    kotlin("jvm") version "1.8.22"
```
TWiStErRob, line 9:
```suggestion
    id("org.jetbrains.kotlin.jvm") version "1.8.22"
```
(renovate)

build.gradle.kts
```diff
@@ -6,7 +6,7 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 
 plugins {
-    `kotlin-dsl`
```
TWiStErRob, line 9: I wonder if we can remove some Kotlin lang version hackery below.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -146,69 +138,74 @@ class NexusPublishPlugin : Plugin<Project> {
...
-            repository.convention(repo)
-            packageGroup.convention(extension.packageGroup)
-            descriptionRegex.convention(extension.repositoryDescription.map { "\\b" + Regex.escape(it) + "(\\s|$)" })
+            it.group = PublishingPlugin.PUBLISH_TASK_GROUP
+            it.description = "Finds the staging repository for ${repo.name}"
+            it.registry.set(registryService)
+            it.usesService(registryService)
+            it.repository.convention(repo)
+            it.packageGroup.convention(extension.packageGroup)
+            it.descriptionRegex.convention(extension.repositoryDescription.map { repoDescription -> "\\b" + Regex.escape(repoDescription) + "(\\s|$)" })
```
TWiStErRob, line 173: `it`'s the reason why I didn't remove it when trying Gradle 8.3

build.gradle.kts
TWiStErRob, line 9: How come it worked so far? Is it a coincidence? luck? or just the warning was there all along?

build.gradle.kts
3flex: Good point - I just did a quick search and it should still work using this notation just FYI for the future: https://github.com/renovatebot/renovate/issues/7777

I'll change it though to be safe not sorry.

build.gradle.kts
3flex: I'm sure that will be possible - maybe in a follow up rather than this PR

3flex: It's worked fine up to now, but when you declared the explicit version for `kotlin-dsl` in #258, that's when it started complaining.

The intention of removing it in this PR is to unblock upgrades to the Gradle version used to build the project.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
3flex: That's an option but IMHO it's safer and potentially less error prone to remove it from the plugin entirely. The syntactical improvements that kotlin-dsl provides are nice, but they're really intended for use by build authors so the Kotlin DSL can be aligned as closely as possible to the older Groovy DSL. I think it's better to remove it and not worry about it. But it's a question of convenience vs compatibility and I'll let a maintainer weigh in.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
3flex: Can do `with(it) {}` if the extra nesting doesn't bother anyone

TWiStErRob: I would stick with `it.` then.

TWiStErRob: Sidenote: we could also apply the SAM compiler plugin manually to get this feature back (I played around in the past, most of the things kotlin-dsl does is possible without it.)

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
TWiStErRob: @3flex so you're saying that kotlin-dsl was meant for buildSrc only?

@szpak thoughts on extra dep to keep these, or accept the Java-compatible syntax and potentially platform types in some places?

szpak: I still write my own scripts in Groovy, so I'm not up to date with the Kotlin stuff :-/.

The syntax is definitely "less pretty". What is the official recommendation of the Gradle team for the plugins? Is kotlin-dsl is really for "buildSrc", not for the custom plugins? Does the Gradle team have a plan to avoid that kind of "compatibility" issues with the next Kotlin versions (such as incoming 2.0)?

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
szpak: > Sidenote: we could also apply the SAM compiler plugin manually to get this feature back (I played around in the past, most of the things kotlin-dsl does is possible without it.)

Are the any side effects/downsides?

TWiStErRob: > Are the any side effects/downsides?

Not sure, I think I abandoned it in the end, because I did something else.
As far as I know it compiles down to the right code in the end, so the only downside I see is that it complicates the build setup of this project, but shouldn't affect consumers.

Probably the best to just go with this PR and then reintroduce pieces separately if we can.

src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
3flex: > What is the official recommendation of the Gradle team for the plugins?

I don't think there are any explicit recommendations regarding use of the Kotlin DSL in published third party plugins.

> Is kotlin-dsl is really for "buildSrc", not for the custom plugins?

It's there to support precompiled script plugins, not binary plugins. But the suggestion to apply `gradleKotlinDsl()` as a dependency instead should avoid the version incompatibility in the plugin while allowing use of the Kotlin extensions, as long as everything is backwards compatible to the oldest supported Gradle version. I'll try that.

> Does the Gradle team have a plan to avoid that kind of "compatibility" issues with the next Kotlin versions (such as incoming 2.0)?

No idea what's on the roadmap for Kotlin DSL.

szpak: @3flex Are you on the Gradle Slack (or their forum)? Maybe it is worth to ask a question about the recommendation in that case? Maybe they have any idea to push it forward (and maybe we would need to switch again to something new soon)?

3flex: Sorry but I'll have to pass on that - my only motivation was to unblock updating Gradle in this project and this PR is a way to do it without breaking anything at the cost of some less pleasant syntax in some places. Removing `kotlin-dsl` completely will definitely not make compatibility worse. Leaving it in might (or might not).

If you'd like to take a different approach I'm happy to make updates, just let me know the direction you'd like to go in, or we can close this PR if you prefer.

TWiStErRob: I think we should go ahead with the PR and make a 2.0 release, and then later on when we drop Gradle 6.x (requiring old Kotlin), we can maybe reintroduce this.

TWiStErRob:
> So shouldn't it be sufficient to just apply the built-in version of `kotlin-dsl` and set the `languageVersion` and / or `apiVersion` to the versions used by the lowest supported Gradle version?

The problem is that when using Gradle 8.3 to compile the nexus plugin the embedded Kotlin compiler (via kotlin-dsl) does not support the 1.3 language version required for the minimum Gradle version supported by nexus plugin. That's a long sentence, here's a ref: https://github.com/gradle/gradle/issues/25868

> Or if these versions are not supported by the Gradle version of this project, either downgrade Gradle

We're here. We're stuck on 8.2.1. This PR tries to unblock this without warnings.

> or force the Kotlin Gradle Plugin to a version where those values are still supported, ignoring the warning that the Kotlin version does not match the embedded version, but still using the built-in `kotlin-dsl` version?

In #258 I think I did this, but maybe it could be done better.

TWiStErRob:
@szpak I'm more and more in favour of just getting this in, to move things along. It's easy to reinstate anything here, after we're on 8.6, or maybe we won't even miss them :) This might even be a non-issue if we drop more Gradle versions post 2.0.

build.gradle.kts
TWiStErRob, line 9: https://github.com/gradle-nexus/publish-plugin/pull/295#issuecomment-1893712584 raises a good point, where is java-gradle-plugin coming from, or we don't need it.

szpak:
> This might even be a non-issue if we drop more Gradle versions post 2.0.

What version would you like to set as a minimum for 2.0?

TWiStErRob:
> What version would you like to set as a minimum for 2.0?

**Post** 2.0 = 2.1+ or 3.0. I think we dropped enough with #274. Let's try to make a release, it's been too long. I have a feeling projects are forced to use 2.0.0-rc-1 if they want full Gradle 8.+ compatibility.

szpak:
> Let's try to make a release, it's been too long.

Fair point.

1. Do we want to merge also #296 before? Or this also is the one you think could be postponed?
2. Would you propose rc-2 to be open for a week or so to gather potential complaints? Or directly "final"?

build.gradle.kts
szpak: Hmm, we use `com.gradle.plugin-publish` which provided `gradlePlugin {}`. We have publications set manually. They could be [some other](https://docs.gradle.org/current/userguide/java_gradle_plugin.html) which could be missing/misconfigured once the plugin is released without java-gradle-plugin.
We could add it - just in case ;-) - or release rc-2 hoping to have it discovered within a week or so.

szpak: Never mind. From the [documentation](https://docs.gradle.org/current/userguide/java_gradle_plugin.html#plugin_publish_plugin):


```
[Plugin Publish Plugin](https://docs.gradle.org/current/userguide/java_gradle_plugin.html#plugin_publish_plugin)

Starting from [version 1.0.0, the Plugin Publish Plugin](https://plugins.gradle.org/plugin/com.gradle.plugin-publish/1.0.0) always auto-applies the Java Gradle Plugin (java-gradle-plugin) and the Maven Publish Plugin (maven-publish).
```

szpak (approved):
@TWiStErRob Feel free to merge it once you think it is ready.

build.gradle.kts
TWiStErRob: Ah, of course...

TWiStErRob:
* This unblocks #296, but but we're still blocked on https://github.com/gradle-nexus/publish-plugin/pull/299#issuecomment-1899362852

PR #296, opened 2024-01-14
Gradle 8.5

3flex:
Closes #277 
Closes #258 

cc @TWiStErRob


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=03ec176d388f2aa99defcadc3ac6adf8dd2bce5145a129659537c0874dea5ad1
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.2.1-bin.zip
+distributionSha256Sum=9d926787066a081739e8200858338b4a69e837c3a821a33aca9db09dd4a41026
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
```
TWiStErRob, line 4: Can you please update stutter matrix so it's also tested? (hopefully it "just works")

TWiStErRob, line 4: Can you please review the new deprecation warnings coming with this upgrade?

TWiStErRob:
@szpak we can include this too in 2.x, because I think it would give us a way to quickly react if there's a compatibility issue.

gradle/wrapper/gradle-wrapper.properties
3flex: I'll take a look

3flex: I had a look at a couple of build scans and only saw deprecation warnings from the stutter plugin which I can't do anything about. The stutter plugin needs to be updated before Gradle 9.0.

Were there any others you're aware of? Can you please point me to the right CI task/build scan?

TWiStErRob: I wasn't, just saw the default deprecation message in the CI logs.

Would you mind opening a stutter issue so that he's aware? There are no open issues/prs and we're on the latest.

TWiStErRob (approved):

PR #298, opened 2024-01-16
Build on latest Java, test through lowest possible - approach 2

szpak:
This is a reworked variant of #254 (I really liked the approach with overriding javaLauncher for test!) with the following changes:
- e2e tests are executed as before with the latest JDK version (to do not put pressure on Sonatype Nexus)
- compatibility tests are executed also with JDK 8 (even if built with higher JDK) - sample [failure](https://github.com/gradle-nexus/publish-plugin/actions/runs/7548032024/job/20549265579#step:6:147) generated on CI on purpose.

This is not as reliable as separate e2eTests executed with lower JDK, but the binaries are built with higher JDK and running tests leverages that binaries.

Downsides:
- no way to use test libraries requiring JDK 11+ (e.g. newer Mockito versions)
- more jobs with compatibility tests on CI
- slightly more complicated stutter configuration (`compatTestJavaauto` added)
- problem with stutter configuration if Gradle drops JDK 8 support (even in toolchain, which is probably not planned yet)
- ?

Maybe you will have an idea how to improve it further.


.github/workflows/java-versions.yml
```diff
@@ -38,4 +39,5 @@ jobs:
       - name: Test cross Java versions compatibility
         run: |
           ./gradlew --version
-          ./gradlew --stacktrace build compatTestJava${{ matrix.jdk }}
+          ./gradlew -q javaToolchains
+          ./gradlew --stacktrace -PnexusPublishPlugin.test.java=${{ matrix.testJdk }} build compatTestJava${{ matrix.jdk }}
```
szpak, line 43: To avoid "javaauto" in stutter, we could add testJdk matrix elements manually, adjusting "auto" version to a given JDK version.
Also a shell command could replace "auto" with a proper version.

However, I don't know which of this 3 options is best :-/

.github/workflows/java-versions.yml
```diff
@@ -15,8 +15,9 @@ jobs:
   openjdk:
     strategy:
       matrix:
-        jdk: [11, 17]
-    name: "OpenJDK ${{ matrix.jdk }}"
+        jdk: [11, 17, 19]
+        testJdk: [auto, 8]
```
szpak, line 19: I wonder if we shouldn't have "fail-fast" disabled for that workflow?

build.gradle.kts
```diff
@@ -261,6 +288,19 @@ tasks {
     withType<Test>().matching { it.name.startsWith("compatTest") }.configureEach {
         systemProperty("plugin.version", project.version)
     }
+    named<Test>("test").configure {
```
3flex, line 291: This is equivalent:
```suggestion
    test {
```

3flex (commented):
Java 21 can be used instead of 19 once #296 is merged.

build.gradle.kts
```diff
@@ -174,7 +190,7 @@ kotlin.target.compilations.configureEach {
     compilerOptions.configure {
         // Gradle fully supports running on Java 8: https://docs.gradle.org/current/userguide/compatibility.html,
         // so we should allow users to do that too.
-        jvmTarget = JvmTarget.fromTarget(JavaVersion.VERSION_1_8.toString())
+        jvmTarget = JvmTarget.fromTarget(project.java.targetCompatibility.toString())
```
3flex, line 193: Consider merging #300 then rebasing.

build.gradle.kts
TWiStErRob: Is that `tasks.test {}` or `project.test {}` just to make sure it's at the right level in the build.gradle.

TWiStErRob:
I'm yet to do a full review, thanks for looking @3flex !

PR #300, opened 2024-01-20
Simplify build file

3flex:
Simplify the build file a bit by updating the way Kotlin API & language versions are declared, and how the JVM target is set.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,6 +5,7 @@ import org.jetbrains.gradle.ext.copyright
 import org.jetbrains.gradle.ext.settings
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 import kotlin.math.min
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
 
 plugins {
     id("org.jetbrains.kotlin.jvm") version "1.8.22"
@@ -169,25 +170,14 @@ sourceSets {
     }
 }
 
+// Supporting Gradle 6.2+ needs to use Kotlin 1.3.
+// See https://docs.gradle.org/current/userguide/compatibility.html
+// For future maintainer: Kotlin 1.9.0 dropped support for Kotlin 1.3, it'll only support 1.4+.
+// This means Gradle 7.0 will be the lowest supportable version for plugins.
+val usedKotlinVersion = @Suppress("DEPRECATION") KotlinVersion.KOTLIN_1_3
+
 kotlin.target.compilations.configureEach {
-    // Supporting Gradle 6.2+ needs to use Kotlin 1.3.
-    // See https://docs.gradle.org/current/userguide/compatibility.html
-    // For future maintainer: Kotlin 1.9.0 dropped support for Kotlin 1.3, it'll only support 1.4+.
-    // This means Gradle 7.0 will be the lowest supportable version for plugins.
-    val usedKotlinVersion = @Suppress("DEPRECATION") KotlinVersion.KOTLIN_1_3
-
-    compilerOptions.configure {
-        // Suppress "Language version 1.3 is deprecated and its support will be removed in a future version of Kotlin".
-        freeCompilerArgs.add("-Xsuppress-version-warnings")
-    }
     compileTaskProvider.configure {
-        // These two (api & lang) needs to be here instead of in compilations.compilerOptions.configure { },
-        // to prevent KotlinDslCompilerPlugins overriding to Kotlin 1.8.
-        compilerOptions.apiVersion = usedKotlinVersion
-        // Theoretically we could use newer language version here,
-        // but sadly the @kotlin.Metadata created on the classes would be incompatible with older consumers.
-        compilerOptions.languageVersion = usedKotlinVersion
-
         // Validate that we're using the right version.
         doFirst {
             val api = compilerOptions.apiVersion.get()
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -198,6 +188,17 @@ kotlin.target.compilations.configureEach {
 }
 
 tasks {
+    withType<KotlinCompilationTask<*>>().configureEach {
```
TWiStErRob, line 169: Why did you move to a global withType, rather than configuring the compilation's options/task?

(Compilations are the way forward, we don't need to know about tasks any more; similar to how AGP's variant and artifact DSL hides the existence of tasks.)

TWiStErRob: As noted in the removed comment (in commit 2) -- now that kotlin-dsl is not messing with us -- we can use the cleanest DSL:

```kotlin
kotlin.target.compilations.configureEach {
  compilations.compilerOptions.configure {
    ...
  }
}
```
No need for any task references.

build.gradle.kts
```diff
@@ -165,34 +165,18 @@ sourceSets {
...
-
-kotlin.target.compilations.configureEach {
-    compileTaskProvider.configure {
-        // Validate that we're using the right version.
-        doFirst {
-            val api = compilerOptions.apiVersion.get()
-            val language = compilerOptions.languageVersion.get()
-            if (api != usedKotlinVersion || language != usedKotlinVersion) {
-                TODO(
-                    "There's mismatch between configured and actual versions:\n" +
```
TWiStErRob, line 198: 🤞 you're right that no-one else is touching this now, neither in the future (as we maintain deps and evolve).

build.gradle.kts
```diff
@@ -99,9 +98,8 @@ dependencies {
     testImplementation(gradleApi())
 }
 
-java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+kotlin {
+    jvmToolchain(8)
```
TWiStErRob, line 103: While this is nicer, it has a big side effect: many machines need to download a separate JDK just to compile 8. Even JDK 21 supports targeting 8!

TWiStErRob: I guess stutter downloads them anyway (when running tests), so it's fine.

TWiStErRob (changes requested):

build.gradle.kts
szpak: > I guess stutter downloads them anyway (when running tests), so it's fine.

However, not everyone runs the tests with stutter for JDK 8. In #298 I propose a concept to build (and run tests) with the current version by default. Only the regression tests require other JDK, but it is handled by CI (or can be triggered by a developer, if needed). In that context the original assignment seems more flexible to a regular developer build.

TWiStErRob: True, we could also argue that since the foojay resolver is pretty common in Gradle builds, the availability of all relevant JDKs is likely on any dev machine doing Gradle-based dev. Even if not, it's an almost negligible one-time cost. If this was actively maintained by hundreds or had cost-sensitive uncached CI, my view would be different.

---

That said, it's probably better to use the default JVM with target 8 as before, so that the target version is not *implicit* from the toolchain. It also reduces any potential contributor's setup complexity.

So the question is what we optimise for? We seem to have a consensus for DX with @szpak. @3flex I'm curious what's are your thoughts on the above points.

build.gradle.kts
3flex: > Why did you move to a global withType

Because [that's what the documentation says to do to](https://kotlinlang.org/docs/gradle-compiler-options.html#target-the-jvm) to modify compiler options. Actually it should be `withType<KotlinJvmCompile>()` because we're not in a multiplatform project, I can update.

The actual cleanest config for setting compiler options would be:
```kotlin
kotlin {
    compilerOptions {
        apiVersion = ...
        languageVersion = ...
    }
}
```

But this requires the project to update to Kotlin 1.9.

There's no mention anywhere [here ](https://kotlinlang.org/docs/gradle-configure-project.html) or [here](https://kotlinlang.org/docs/gradle-compiler-options.html) of the syntax that's currently used or what you proposed above?

build.gradle.kts
3flex: > so that the target version is not implicit from the toolchain

That's a feature, not a bug! And I don't think "setup complexity" is an issue here, the whole point of toolchains is that a user doesn't have to muck around with JDKs when toolchains are in use. It's a one-off download if JDK 8 isn't already on the machine. After that things "just work".

Toolchains are prominently featured in the Kotlin docs for configuring a Gradle project so I don't think this config will surprise anyone.

It's not hard to remove use of the toolchain here though if it's not wanted.

szpak: @3flex What is a "problem" with using the current approach with building the project (by default) with the JDK that is used to start Gradle?

Support for building with JDK 8 is [deprecated](https://github.com/gradle-nexus/publish-plugin/pull/303), so I don't know, if forcing users to have (or automatically download JDK 8) matches the aforementioned.

3flex: It's not a problem. It just makes the config simpler and in line with best practice.

I'm going to remove the use of toolchains then to get this over the line.

build.gradle.kts
3flex: This was the last blocking bit of feedback which I believe I've answered above. Are you happy to approve?

TWiStErRob (approved):

PR #301, opened 2024-01-20
Remove buildSrc

3flex:
Only one function in buildSrc was actually required in the build, and moving that into build.gradle.kts was simple. Removing buildSrc will speed up build configuration.


build.gradle.kts
```diff
@@ -43,6 +46,10 @@ repositories {
 }
 
 val licenseHeaderFile = file("gradle/license-header.txt")
+fun readCopyrightHeader(licenseHeaderFile: File) =
+    licenseHeaderFile.readLines()
+        .joinToString("\n") { line -> line.substring(min(line.length, 3)) }
```
TWiStErRob, line 51: Probably not in scope, but could this be simplified with one of the many `trimIndent` and friends variants?

TWiStErRob (approved):

TWiStErRob:
Really nice catch 💪💡

TWiStErRob:
Thank you for contributing these PRs! Things are getting very clean.

build.gradle.kts
TWiStErRob: Merging this PR, I think this could be in scope for "simplify build.gradle"

szpak:
>  Removing buildSrc will speed up build configuration.

Definitely, especially as it is a configuration in Kotlin... Thanks.

PR #308, opened 2024-01-25
Note how to override default identity values

Goooler:


TWiStErRob:
I don't think this is necessary. In a pure JAR publication the project name should be the artifact name.

Goooler:
For those who are new to Maven publishing, they may not have a clear understanding of how to modify the artifactId.

TWiStErRob:
https://docs.gradle.org/current/userguide/publishing_maven.html#sec:identity_values_in_the_generated_pom

TWiStErRob:
The defaults are there, and it also shows how to override. I would agree to add a link to this, but not to add it the examples, because the default values are good most of the time, and if someone wants to override, they should find this page somehow for their _advanced_ use.

README.md
```diff
@@ -154,6 +154,9 @@ nexusPublishing {
 ```
 
 #### Add Metadata ####
+
+Override the default identity values with [this guide](https://docs.gradle.org/current/userguide/publishing_maven.html#sec:identity_values_in_the_generated_pom).
```
TWiStErRob, line 158:
```suggestion
See the [Maven publishing page in Gradle documentation](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:publications) for what these mean, default values, and how to override coordinates (`groupId:artifactId:version`).
```

TWiStErRob (approved):

Issue #320, opened 2024-02-08
Support Nexus Pro 3


sradi:
Hi,

`io.github.gradle-nexus.publish-plugin` fails with http 404, when I try to use it with Nexus Pro 3:

```
Execution failed for task ':initializeReleaseStagingRepository'.
> Failed to load staging profiles, server at https://<nexusProUrl>/repository/<repoName/ responded with status code 404, body:
  <!DOCTYPE html>
  <html lang="en">
  <head>
    <title>404 - Sonatype Nexus Repository</title>
...
```

The REST API of Nexus Pro 3 (3.63.0-01) doesn't seem to offer any endpoint for `stagingProfiles`. Instead, one is supposed to tag artifacts and stage them by moving them to another repository (`/v1/staging/move/{destination}?tag=...`).

Do I assume correctly, that this Plugin is only compatible with Nexus Pro 2.x?
If this plugin could offer to add a tag to each published artifact, this would help a lot for staging or dropping a set of artifacts.

szpak:
Nexus 3 is rewritten a piece of software from scratch with completely different API. This plugin supports only Nexus 2 which powers The Central Repository (aka Maven Central). I also haven't heard about the Sonatype plans to switch publishing to TCR to Nexus 3. The main use case it to publish artifacts to TCR.

> If this plugin could offer to add a tag to each published artifact, this would help a lot for staging or dropping a set of artifacts.

It looks like the completely new API. I would be probably hard to maintain both APIs (and related workflows) in the same plugin.

Let's keep this request open, but probably it would be easier to write a separate plugin for that (but I haven't analized the problem in depth).

sradi:
Thanks for the clarification. I wasn't aware, that Nexus 3 is a completely rewritten Software.

I totally agree, that it doesn't make sense to squeeze support for both Nexus versions into this plugin.
What do you think about making the docs clearer on, what Nexus versions are supported? 

In case, the Sonatype Enterprise Support shares some best practices with me, how to publish to Nexus Pro 3 from Gradle, I am going to share it here.

szpak:
> What do you think about making the docs clearer on, what Nexus versions are supported?

Good idea. I believe, you are the second person with the similar problem. Would you like to create a PR with the proposed change?

> In case, the Sonatype Enterprise Support shares some best practices with me, how to publish to Nexus Pro 3 from Gradle, I am going to share it here.

:+1:

PR #321, opened 2024-02-11
Update docs on supported Nexus versions

sradi:
Clarification, that Nexus 3 is not supported, as suggested in #320.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![CI Status](https://github.com/gradle-nexus/publish-plugin/workflows/CI/badge.svg)](https://github.com/gradle-nexus/publish-plugin/actions?workflow=CI) [![Gradle Plugin Portal](https://img.shields.io/maven-metadata/v/https/plugins.gradle.org/m2/io.github.gradle-nexus/publish-plugin/maven-metadata.xml.svg?label=Gradle%20Plugin%20Portal)](https://plugins.gradle.org/plugin/io.github.gradle-nexus.publish-plugin)
 
-This Gradle plugin is a turn-key solution for publishing to Nexus. You can use it to publish your artifacts to any Nexus Repository Manager 2.x instance (internal or public). It is great for publishing your open source to Sonatype, and then to Maven Central, in a fully automated fashion.
+This Gradle plugin is a turn-key solution for publishing to Nexus. You can use it to publish your artifacts to any Nexus Repository Manager 2.x instance (internal or public). It is great for publishing your open source to Sonatype, and then to [Maven Central][maven-central], in a fully automated fashion.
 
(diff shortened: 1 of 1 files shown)


szpak (approved):

README.md
```diff
@@ -2,14 +2,17 @@
...
 
 Vanilla Gradle is great, but it cannot fully automate publications to Nexus. This plugin enables isolation of staging repositories so that you can reliably publish from CI, and each publication uses a brand new, explicitly created staging repo ([more](https://github.com/gradle-nexus/publish-plugin/issues/63)). Moreover, the plugin provides tasks to close and release staging repositories, covering the whole releasing process to Maven Central.
 
 This plugin is intended as a replacement of the [Gradle Nexus Staging Plugin](https://github.com/Codearte/gradle-nexus-staging-plugin/) and [Nexus Publish Plugin](https://github.com/marcphilipp/nexus-publish-plugin) duo. See a dedicated [migration guide](https://github.com/gradle-nexus/publish-plugin/wiki/Migration-from-gradle_nexus_staging-plugin---nexus_publish-plugin-duo).
 
 ## Usage
 
+> [!NOTE]  
+> Nexus Repository Manager 3 is not supported. The staging concept and REST API between 2.x and 3.x totally differ.
```
TWiStErRob, line 14: I think this would be confusing to most users. e.g. I've never heard about this before. As a first time "I just want to publish my jar" integrator, seeing a "blah not supported" as the first sentence in "how to use this thing?" section would scare me.

sradi: What do you propose instead?

TWiStErRob: I would go with an h3 "Compatibility" section, probably right before Troubleshooting.

You could create an h4 for Nexus. And I think we should later add Java, Gradle, Kotlin sections at minimum. Or just a bullet list instead of h4. Depends on how much detail.

szpak: @sradi What do you think about tweaks proposed by Rob?

sradi: Sounds good. Maybe h4 or bullet list can be added, when there are more than just Nexus compatibility details added in that section.

TWiStErRob: Nice, I updated it a bit (sorry, couldn't suggest because it was complex), @szpak I think it's good to go, please have a [glance](https://github.com/sradi/publish-plugin/blob/nexus3-docs-improvement/README.md#compatibility).

TWiStErRob (approved):

README.md
szpak: Look good, thanks.

Btw, in the next iteration, we would need to re-think and document also (our intentions for) support for https://github.com/gradle-nexus/publish-plugin/issues/330

PR #328, opened 2024-04-05
Migrate to develocity plugin

Goooler:
https://docs.gradle.com/enterprise/gradle-plugin/legacy/#develocity_migration

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -7,6 +7,7 @@ develocity {
     buildScan {
         termsOfUseUrl = "https://gradle.com/help/legal-terms-of-use"
         termsOfUseAgree = "yes"
+        uploadInBackground = false
         publishing.onlyIf { System.getenv("CI") != null }
     }
 }
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -7,9 +7,7 @@ develocity {
     buildScan {
         termsOfUseUrl = "https://gradle.com/terms-of-service"
         termsOfUseAgree = "yes"
-        // TODO: workaround for https://github.com/gradle/gradle/issues/22879.
-        val isCI = providers.environmentVariable("CI").isPresent
-        publishing.onlyIf { isCI }
+        publishing.onlyIf { System.getenv("CI") != null }
     }
 }
 
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,15 +1,15 @@
 plugins {
-    id("com.gradle.enterprise") version "3.17"
+    id("com.gradle.develocity") version "3.17"
     id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
 }
 
-val isCiBuild = System.getenv("CI") != null
-
-gradleEnterprise {
+develocity {
     buildScan {
-        termsOfServiceUrl = "https://gradle.com/terms-of-service"
-        termsOfServiceAgree = "yes"
-        publishAlwaysIf(isCiBuild)
+        termsOfUseUrl = "https://gradle.com/terms-of-service"
+        termsOfUseAgree = "yes"
+        // TODO: workaround for https://github.com/gradle/gradle/issues/22879.
+        val isCI = providers.environmentVariable("CI").isPresent
+        publishing.onlyIf { isCI }
     }
 }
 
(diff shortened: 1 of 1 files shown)


settings.gradle.kts
```diff
@@ -1,15 +1,15 @@
...
 
-val isCiBuild = System.getenv("CI") != null
-
-gradleEnterprise {
+develocity {
     buildScan {
-        termsOfServiceUrl = "https://gradle.com/terms-of-service"
-        termsOfServiceAgree = "yes"
-        publishAlwaysIf(isCiBuild)
+        termsOfUseUrl = "https://gradle.com/terms-of-service"
```
TWiStErRob, line 8: hmm, did the url change?

https://docs.gradle.com/enterprise/gradle-plugin/current/#connecting_to_scans_gradle_com

or is this something else?

settings.gradle.kts
```diff
@@ -1,15 +1,15 @@
...
+        termsOfUseAgree = "yes"
+        // TODO: workaround for https://github.com/gradle/gradle/issues/22879.
+        val isCI = providers.environmentVariable("CI").isPresent
+        publishing.onlyIf { isCI }
```
TWiStErRob, line 12: They have exactly the same example here:

https://docs.gradle.com/enterprise/gradle-plugin/current/#publishing_based_on_criteria

Could we stick with `getenv` as they show? I agree providers are the way, but if not working, then I would stick with the old way.

is it possible there may be a different syntax, something like:

```
publishing.onlyIf(providers.environmentVariable("CI").orElse(false))
```

settings.gradle.kts
Goooler: `termsOfServiceUrl` has been replaced by `termsOfUseUrl`.

settings.gradle.kts
Goooler: It can't be inline, this is why I commented TODO here.

settings.gradle.kts
TWiStErRob: I meant the value of it, see the docs.

TWiStErRob: It looks like both are accepted for now.

settings.gradle.kts
Goooler: Inlining `!System.getenv("CI").isNullOrEmpty()` is working but `providers.environmentVariable("CI").isPresent`, I can use the former if you prefer.

settings.gradle.kts
TWiStErRob: Let's use the new one:
```suggestion
        termsOfUseUrl = "https://gradle.com/help/legal-terms-of-use"
```
It's in the develocity plugin docs + if you make a typo in it, it'll recommend this one.

settings.gradle.kts
```diff
@@ -1,15 +1,13 @@
...
-
-gradleEnterprise {
+develocity {
     buildScan {
-        termsOfServiceUrl = "https://gradle.com/terms-of-service"
-        termsOfServiceAgree = "yes"
-        publishAlwaysIf(isCiBuild)
+        termsOfUseUrl = "https://gradle.com/help/legal-terms-of-use"
+        termsOfUseAgree = "yes"
+        publishing.onlyIf { System.getenv("CI") != null }
```
TWiStErRob, line 10:
```suggestion
        uploadInBackground = false
        publishing.onlyIf { System.getenv("CI") != null }
```

I think we need this based on:
https://docs.gradle.com/enterprise/gradle-plugin/current/#configuring_background_uploading

And I think this is not related to the migration, it should have been on before too (same docs in legacy too). @szpak did you ever have a problem where a scan wasn't available?

TWiStErRob (approved):
@Goooler Looks good, thank you for looking and doing it!

@szpak please have a look at the scope creep, or merge and we can discuss a separate PR.

settings.gradle.kts
szpak: I don't recall that situation, however, I don't often use build scans, so I might miss it.

szpak: Maybe because there is some time consumed do save local cache. Probably we should set it to `false`, as you suggested.

szpak (approved):

Issue #81, opened 2021-03-20
Isolation in multi-project build


flelli:
Hi,

I have a polyglot multi-project build where sub projects are isolated by technology. Say I have 2 subprojects:

- `modules:go`
- `modules:java`

The reason I do this is also because I want the tools and plugins to be isolated. With the former 2 plugins I had them applied only in the `modules:java` along with Maven, the publishing, etc but the new one fails to apply if not declared in the root project:

`Failed to apply plugin 'io.github.gradle-nexus.publish-plugin'.
   > Plugin must be applied to the root project but was applied to :modules:java`

Moving the plugin to the root project also implies declaring all the publishing etc in the root project and that would mess up all the configurations I have, basically giving up on isolation.

Is there a way to keep the new plugin in a sub-project?

marcphilipp:
> Moving the plugin to the root project also implies declaring all the publishing etc in the root project and that would mess up all the configurations I have, basically giving up on isolation.

That's not entirely true. You only need to configure `nexusPublishing` in the root project but can continue declaring your publications in `modules:java`.

> Is there a way to keep the new plugin in a sub-project?

I'm afraid not. The alternative to a root project plugin would be to convert it into a settings plugin but that wouldn't achieve isolation, either.

flelli:
Thanks Marc for your prompt reply,

The point is if I move `nexusPublishing` only, the build breaks as it can't find any Maven publications defined. So I have to move them as well, which will likely need the signing to be moved too, ending up with moving it all.

> > Moving the plugin to the root project also implies declaring all the publishing etc in the root project and that would mess up all the configurations I have, basically giving up on isolation.
> 
> That's not entirely true. You only need to configure `nexusPublishing` in the root project but can continue declaring your publications in `modules:java`.

marcphilipp:
@flelli It shouldn't break without any defined publications. Could you please create a reproducer project?

flelli:
Sure. In [my project](https://github.com/mooltiverse/nyx) I created two additional branches:

* [demo/ok](https://github.com/mooltiverse/nyx/tree/demo/ok): with the current configuration, using the former 2 separate plugins isolated in the `modules/java/build.gradle`
* [demo/ko](https://github.com/mooltiverse/nyx/tree/demo/ko): with the new plugin in the root project

As soon as I run Gradle using the plugin in the root project, I get:
```
[...]

PR #359, opened 2025-01-12
Allow applying the plugin to a non-root project

andrewaylett:
This introduces the idea of `independentProjects`, defaulting to false, and limiting application of the plugin to just the project applied to if it's true.

By itself, this change should allow projects with a single subproject (such as is templated by `gradle init`) to apply the plugin to that project only.  It's also laying a little ground-work for cross-project synchronisation of staging repositories that's compatible with isolated projects.

This PR is in response to #81 but is not a complete solution.

### DIFF
--- a/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
+++ b/src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
@@ -43,9 +43,7 @@ class NexusPublishPlugin : Plugin<Project> {
     }
 
     override fun apply(project: Project) {
-        require(project == project.rootProject) {
-            "Plugin must be applied to the root project but was applied to ${project.path}"
-        }
+        val isRoot = project == project.rootProject
 
         require(GradleVersion.current() >= GradleVersion.version("6.2")) {
             "io.github.gradle-nexus.publish-plugin requires Gradle version 6.2+"
@@ -55,7 +53,7 @@ class NexusPublishPlugin : Plugin<Project> {
         val extension = project.extensions.create(NexusPublishExtension.NAME, NexusPublishExtension::class.java)
         configureExtension(project, extension)
         configureNexusTasks(project, extension, registry)
-        configurePublishingForAllProjects(project, extension, registry)
+        configurePublishingForAllProjects(project, extension, registry, isRoot)
     }
 
     private fun configureExtension(project: Project, extension: NexusPublishExtension) {
@@ -68,6 +66,7 @@ class NexusPublishPlugin : Plugin<Project> {
             connectTimeout.convention(Duration.ofMinutes(5))
             transitionCheckOptions.maxRetries.convention(60)
             transitionCheckOptions.delayBetween.convention(Duration.ofSeconds(10))
+            independentProjects.convention(false)
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/github/gradlenexus/publishplugin/NexusPublishPlugin.kt
```diff
@@ -43,9 +43,7 @@ class NexusPublishPlugin : Plugin<Project> {
     }
 
     override fun apply(project: Project) {
-        require(project == project.rootProject) {
-            "Plugin must be applied to the root project but was applied to ${project.path}"
-        }
+        val isRoot = project == project.rootProject
```
TWiStErRob, line 46: I wondering if this change alone is enough? If the user applies to non-root, then it implicitly start using the independent mode? Since it was not allowed before, it means we can define any behaviour for that case.

andrewaylett: Up to you -- I didn't want to break any of your existing tests.

Also, I'm not sure what would happen if someone tried applying the plugin to multiple projects in a build.  I suspect it won't work properly; I'm pretty sure it'll open multiple staging repositories.  So I wonder if needing to specify an "incubating" property is useful as an indication that the behaviour might not be entirely stable?

szpak: Having the plugin applied multiple times in the project might sometimes be useful for monorepos which (for some reason) would like to release particular projects to different locations in Central. However, the project author would then need to ensure that group/project/version is property set in that (sub)project and probably some other element which the plugin gets from the project.

Looking at some comments in #81:
>> Moving the plugin to the root project also implies declaring all the publishing etc in the root project and that would mess up all the configurations I have, basically giving up on isolation.

> That's not entirely true. You only need to configure nexusPublishing in the root project but can continue declaring your publications in modules:java.

And:
https://github.com/gradle-nexus/publish-plugin/issues/81#issuecomment-2049665713

> I originally made it a root project plugin so there's a single place to configure Nexus repositories that are shared across subprojects. While each subproject has an "init" task, they use a resource attached to the project to synchronize on to ensure a shared staging repo is created. This could probably be changed to create a checksum of the repo config or require some kind of identifier and a BuildService but I lack the time to give it a try.

I'm a little afraid to add that switch and later on have an another set of issues related to "corner cases" :thinking: 

I have no experience with [build_services](https://docs.gradle.org/current/userguide/build_services.html). Do you? How hard would it be to convert the plugin to it? Would it "fix" all the (related) problems?

andrewaylett: I've been pondering the question of how I'd go about turning this into a build service -- I do have some experience with them, but not of the degree of complexity that this plugin might require if we wanted to minimise surprises for existing users who upgrade.

There are some design questions that we'd need to decide on answers for.  I'm not sure I have all the questions yet, let alone the answers.

If you'd be interested in seeing the result of my attempt to refactor a bit more thoroughly (with absolutely no obligation to accept any of my code, or indeed any of the design choices I might make) then I'll probably quite enjoy the diversion.  I'm yak shaving for a personal project, interesting problems are if anything more what I'm looking for than actual solutions :).

PR #364, opened 2025-03-18
Stick to Git hash for GH actions

szpak:
See:
https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-pre-written-building-blocks-in-your-workflow#using-shas


TWiStErRob:
This opened https://github.com/gradle-nexus/publish-plugin/pull/365

PR #377, opened 2025-05-06
Add note about Sonatype Central to README.md

snazy:

PR #384, opened 2025-05-21
Legacy OSSRH tokens are incompatible with the new central portal

bric3:
With the sunset of OSSRH and the migration to central, I found out not clear that one need to actually generate new central token as well, i.e. the previous OSSRH generated credentials (from nexus) won't work (results in `401 Unauthorized`)

* https://central.sonatype.org/news/20250326_ossrh_sunset/
* #377

This PR just put more weight on this point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ nexusPublishing {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


szpak (approved):

szpak:
Thanks Brice!

PR #388, opened 2025-06-06
Remove duplicated sonatype repo definition in README

breedx-splk:
Given the upcoming OSSRH EOL/sunset and the number of users that will be migrating, it seems that this block can cause confusion. The `sonatype` definition is present twice, and will cause builds to fail with the follow error:

```
* What went wrong:
Cannot add a NexusRepository with name 'sonatype' as a NexusRepository with that name already exists.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,6 @@ In order to publish to Maven Central (aka the Central Repository or just Central
 ```groovy
 nexusPublishing {
     repositories {
-        sonatype()
-    }
-    repositories {
         // see https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#configuration
         sonatype {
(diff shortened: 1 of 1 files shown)


TWiStErRob (approved):

README.md
```diff
@@ -35,9 +35,6 @@ In order to publish to Maven Central (aka the Central Repository or just Central
 
 ```groovy
 nexusPublishing {
-    repositories {
-        sonatype()
-    }
```
TWiStErRob, line 40: @szpak I wonder if this method should be deprecated or the URLs changed to the new ones.

dr0i:
Why is this not merged?

